PR #3, opened 2020-01-10
add measure and note index to note object

kittyshi:


mgrachten:
Thanks for this PR. I understand the need to refer to the notes by their measure/note index in the document order, to be able to use partitura in combination with Nakamura's alignment tool. That is something currently not possible in partitura. I'm not sure however that the best solution is to hard-code the note and measure indices into the notes. The two issues I see with the current PR:

1. The number attribute of the measure object is obtained from the xml. Although the number attribute of the `<measure>` element is required, it is not guaranteed that the number attributes coincide with the order of the measures, nor that they are unique. I mean you could have a valid (if unlikely) XML file containing:

```xml
   <measure number="2">...</measure>
   <measure number="2">...</measure>
   <measure number="1">...</measure>
```

This PR would give you incorrect values for the `measure_index` attribute, assuming that Nakamura's code simply uses the count of the measurement elements (i.e. the document order) as the index.

2. The `note_index` attribute is rather specific to the use case of the Nakamura alignment, especially since it restarts counting within every measure. The optional "id" attribute of xml `<note>` elements would be ideally suited to refer to xml notes in the score--performance alignment, but I understand that the alignment tool should also be able to deal with xml files where notes do not have id attributes.

My proposed solution to work with alignments from Nakamura's tool would be the following:

1. Once the xml file is loaded, you can get the measure_index values per measure simply like this:

   ```python
   import partitura
   from partitura.score import Measure
   part = partitura.load_musicxml(xml_file)
   measure_indices = [i+1 for i, _ in enumerate(part.iter_all(Measure))]
   ```
   
2. The function `load_musicxml` has an option `force_note_ids` that ensures that every note has a unique id. Currently this function does not guarantee anything beyond the uniqueness of the ids, but I can change it such that the lexicographical order of the ids it produces always coincides with the document order of the notes. Once that is in place, you would produce the `measure_index`/`note_index` attributes like this:

   ```python
   import partitura
   from partitura.score import Measure, Note
   
   part = partitura.load_musicxml(args.file, force_note_ids=True)
   
   for i, m in enumerate(part.iter_all(Measure)):
   
       notes = sorted(part.iter_all(Note, m.start, m.end),
                      key=attrgetter('id'))
   
       for j, n in enumerate(notes):
   
           n.measure_index = i+1
           n.note_index = j+1
    ```

kittyshi:
Hi Maarten, Thanks for the explain. 
However, the above code won't give the same measure and index number as Nakamura's matching.
Partly because the matching uses absolute order as it appears in the xml file. 
I understand that my method would not incorporate edge cases. Currently I wrote for my own use and it works for most of the case. So feel free to neglect.

neosatrapahereje:
Hi @kittyshi and @mgrachten! I think it is important to include a way of encoding the **document order**, not only for MusicXML but for the future MEI support (we have a new student who is working on that right now). Furthermore, I think it is important to add *native support* for Nakamura's match format (i.e., a method which takes a matchfile + a score (MusicXML or MIDI) + the performance MIDI and returns a `PerformedPart` object). Therefore, I would suggest to add a way to code the order in which the notes appear in the MusicXML directly in the constructor of the `Note` or `GenericNote` objects (perhaps add a `do_ix` property, and if none is provided, assign one directly when making an instance of `Score`?)

mgrachten:
@neosatrapahereje the document order is now preserved as part of the Nakamura file format support in `develop`, isn't it? If so, then I'll close this PR.

neosatrapahereje:
Yes @mgrachten, the document order is now preserved (in attribute `do_idx` in `GenericNote` objects). So in the case of your example  in https://github.com/OFAI/partitura/pull/3#issuecomment-573093160, the code would be:

```python
import partitura
from partitura.score import Measure, Note

part = partitura.load_musicxml(args.file, force_note_ids=True)

for i, m in enumerate(part.iter_all(Measure)):

    notes = sorted(part.iter_all(Note, m.start, m.end),
                   key=attrgetter('do_idx'))

    for j, n in enumerate(notes):

        n.measure_index = i+1
        n.note_index = j+1
```

That being said, there is now a method for importing files from Nakamura's alignment tool in branch `nakamura_matchformat`, which will be merged in develop soon.

mgrachten:
Ok, then I'll close this PR.

Issue #13, opened 2020-12-11
Examples for methods in `partitura.musicanalysis`


neosatrapahereje:
It would be important to add examples in the documentation to all methods in `partitura.musicanalysis` module. This would make the usage of these methods clearer to the users without having to dive into the code.

Methods with missing examples:

* `estimate_spelling` (in `partitura.musicanalysis.pitch_spelling`)
* `estimate_key` (in `partitura.musicanalysis.key_identification`)
* `estimate_voices` (in `partitura.musicanalysis.voice_separation`)
* `estimate_tonaltension`(in `partitura.musicanalysis.tonaltension`)

mgrachten:
I merged my local commits into `develop`, which add an example for `estimate_key` to `docs/usages.rst`, in a section `music analysis`.  This section can be extended with example calls for the other functions.

neosatrapahereje:
Commit 7b13650 adds the examples in the documentation, and are now included in pull request #26

PR #26, opened 2021-03-23
Update api musicanalysis

neosatrapahereje:
All methods in `partitura.musicanalysis` work well with the new `note_array`s, and it is now possible to use the methods for both `Part`, `PartGroup`, `PerformedPart` objects, as well as lists of `Part` objects.

There are two new methods to get structured `note_array`s from `Part`s, `PartGroup`s or lists of `Part`s: `note_array_from_part_list` and `note_array_from_part`.  These methods are now used in the `note_array` properties of `Part` and `PartGroup` objects to avoid unnecessary code duplication.


neosatrapahereje:
Commit  7b13650 adds the example of usage of music analysis methods from issue #13

Issue #27, opened 2021-03-25
Update dtypes


neosatrapahereje:
As of Numpy 1.20.0, the use of `np.int` or `np.float` is deprecated, and it is suggested to use `int` or `float` by itself. Otherwise, we get the following warnings:

```
DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

```
DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

mgrachten:
Yes, we should update the types.

PR #33, opened 2021-04-15
Update dtypes

neosatrapahereje:
This pull request solves issue #27.


mgrachten:
Although defining the int and float type globally is correct, to be honest I think I would prefer to keep it simple and just replace `np.int` by `int` and `np.float` by `float`.

Issue #31, opened 2021-04-13
The method musicxml_to_notearray does not work on scores with multiple parts


melkisedeath:
The Error occured when tried to parse a score with 4 parts and the number of note instances in every part are different.
```
Traceback (most recent call last):
  File "src\descriptors\note_density.py", line 132, in <module>
    note_array = partitura.musicxml_to_notearray(my_musicxml_file)
  File "C:\Users\melki\Desktop\JKU\codes\Music-Structure-Segmentation\venv\lib\site-packages\partitura\io\importmusicxml.py", line 1304, in musicxml_to_notearray
    scr = np.vstack(scr)
  File "<__array_function__ internals>", line 5, in vstack
  File "C:\Users\melki\Desktop\JKU\codes\Music-Structure-Segmentation\venv\lib\site-packages\numpy\core\shape_base.py", line 283, in vstack
    return _nx.concatenate(arrs, 0)
  File "<__array_function__ internals>", line 5, in concatenate
ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 300 and the array at index 1 has size 257
```
The method musicxml_to_notearray is quite old, and should be deprecated. It was created in an early version of partitura, but then we moved to a different way to get note_arrays.

It should be replaced by the method ensure_notearray (in partitura.utils).

Example : 
```
note_array = partitura.utils.ensure_notearray(partitura.load_musicxml("/my/score.musicxml"))
```

PR #34, opened 2021-04-15
Fix issue #31

neosatrapahereje:
This pull request includes 3 things:

1. Updates `musicxml_to_notearray` to work with the new fields in `note_array`s. The method is now analogous to `midi_to_notearray`, i.e., a simple and convenient way to load a MusicXML file into an array. This fixes the main part of issue #31. This is also addressed in commit e9d8439.
2. Fixed a bug when getting a `note_array` when no voice information is given (to avoid errors with `None` being casted as an integer). This was the specific problem with the MusicXML file that motivated the issue. This is also addressed in commit 
e9d8439.
3. Fixed a bug when creating `Part`s from MatchFiles (which was one of the motivations of the the original discussion @melkisedeath and @nimrodVarga that motivated the Issue). This is addressed in commit d47ba42.

### DIFF
--- a/partitura/io/exportmidi.py
+++ b/partitura/io/exportmidi.py
@@ -7,7 +7,7 @@ from collections import defaultdict, OrderedDict
 from mido import MidiFile, MidiTrack, Message, MetaMessage
 
 import partitura.score as score
-from partitura.utils import partition, INT_TYPE, FLOAT_TYPE
+from partitura.utils import partition
 
 __all__ = ["save_score_midi", "save_performance_midi"]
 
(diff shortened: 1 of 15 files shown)


neosatrapahereje:
There have been a few updates since the opening the pull request:

1. In commit ea67858, @sildater added a more permanent fix for incongruent onset time information (in anacrusis) when creating a `Part` from a match file (c.f., with commit d47ba42)
2. Adds an option to `unfold_part_maximal` to update the note IDs to reflect the repetition structure. The updated note IDs follow the conventions from the Magaloff/Zeilinger dataset (i.e., `n123-1` and `n123-2` are the first and second repetitions of note `n123`). 
3. Adds a method for unfold a part given an alignment (`unfold_part_alignment`). The method relies on make_score_variants, and thus, does not currently support nested repeats, Da Capo/Fine/Segno signs.

The last two points were addressed in commit 832de2f.

PR #36, opened 2021-04-23
Update unfold part

neosatrapahereje:
This branch has 2 contributions:

1. Adds an option to `unfold_part_maximal` to update the note IDs to reflect the repetition structure. The updated note IDs follow the conventions from the Magaloff/Zeilinger dataset (i.e., `n123-1` and `n123-2` are the first and second repetitions of note `n123`)
2. Adds a method for unfold a part given an alignment (`unfold_part_alignment`). The method relies on `make_score_variants`, and thus, does not currently support nested repeats, Da Capo/Fine etc.

### DIFF
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -2431,6 +2431,16 @@ def unfold_part_maximal(part, update_ids=True):
     part where all segments marked with repeat signs are included
     twice.
 
+    Parameters
+    ----------
+    part : :class:`Part`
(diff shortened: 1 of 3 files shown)


mgrachten:
This PR breaks beckwards compatibility by setting the default of ID update option to `True`. If you think this is a better default (do you?) then let's do it, but in that case we need to look at the unfolding test, which is fails in this PR.

PR #42, opened 2021-05-04
Nakamura test

mgrachten:
Proof of concept `np.loadtxt`-based support for Nakamura corresp files. I guess the same could be done for the Nakamura match format.


neosatrapahereje:
The new `np.loadtxt`-based loader for Nakamura's match format was added, including documentation and tests.

PR #46, opened 2021-09-17
Metrical grid

fosfrancesco:
A new field is added to the note_array, that identifies the position of the downbeat and the time elapsed since the latest downbeat (i.e. the position in the measure).
All the information related to beat can now be also computed using "musical beat" as the reference. For example, a measure in 6/8 will be considered with 2 beats instead of 6.


neosatrapahereje:
Hi @fosfrancesco! 

The changes look quite good.  

Perhaps it could be useful to add a flag (e.g., `use_musical_beat` to `Part` objects so that the `note_array` property can use either the "notation" beats (would this be the right term?) or the new musical beats? Something like 

```python
    @property
    def note_array(self):
        return note_array_from_part(self,  use_musical_beat=self.use_musical_beat)
```

We could also add this option to all methods for loading scores (i.e., `load_musicxml`, `load_score_midi`, etc.).
What do you think about that @mgrachten and @sildater? 

Also, we should include an explanation of this behavior and the difference between notation and musical beats in the documentation.

Thanks for your first contribution to Partitura! 😄 

Best,

Carlos

fosfrancesco:
Thank you! The "musical beat" is the default choice in other libraries like PrettyMidi and music21. There it is just called "beat", as opposed to "time signature numerator". But in order to make it compatible with the previous version, we can keep the names "musical beat" and "notation beat".

And sure, I can easily add a field in part if you confirm that this is something that we want.

Francesco

fosfrancesco:
I put `use_musical_beat` as a field in the Part object. I didn't add the option in the XML import because I felt it was a very specific parameter for a generic function. 

Instead, the attribute can be set True by using the new method `set_musical_beat()`. 
This function accepts a parameter to set a custom number of beats (of equal duration) for a given time signature.
For example, to set 6/8 with 3 beats and 2/4 with 1 beat, the parameter of the function will be the dictionary `{"6/8":3, "2/4":1}`.
If no parameter is passed, the default number of musical beats is considered.

Time signatures with beats of different lengths (e.g. 5/4 counted as 3+2) and the option to have the same time signature with different musical beats (e.g. a measure of 6/8 with 2 and another measure with 6) are still not supported.

@neosatrapahereje @sildater @mgrachten  let me know what you think about it

Francesco

neosatrapahereje:
Yes, I like the changes, and that they don't change the default behavior.

> I put use_musical_beat as a field in the Part object. I didn't add the option in the XML import because I felt it was a very specific parameter for a generic function.

Yes, after thinking about it over the weekend, I agree that it would be just easier to handle everything from the `Part` objects. 

In any case, we should add an explanation and an example in the documentation of how/when to use musical beats vs. notated beats.  I'll create a branch for the new updates/additions to the documentation for the next version (0.5.0).

> Time signatures with beats of different lengths (e.g. 5/4 counted as 3+2) and the option to have the same time signature with different musical beats (e.g. a measure of 6/8 with 2 and another measure with 6) are still not supported.

That could be a good addition, but we can save it for the future 😎

I will merge the pull request tomorrow, in case @sildater and @mgrachten have any comments.

Cheers,

Carlos

sildater:
For the last point I refer to something like the first measure of Chopin's Op. 38 (as in the Vienna 4x22 excerpt) where rel_onset_div currently starts with zero at the first note which is also considered a downbeat. I'm wondering whether this is intended behavior because I wouldn't necessary call it a downbeat. What do you think?

fosfrancesco:
Good points!
The initial reason for having the `use_musical_beat` in part objects, was for people to be able to obtain the representation in musical beats when using the default note-array. But since this is not the default choice and has to be set, probably it is better to go back to have it only on the `note_array_from_part`, in order to the setting/unsetting part that can be confusing. @neosatrapahereje let me know if you agree with that since you were the one suggesting the modification.

The solution to avoid float beats value would be to return a `fractions.Fraction` object instead of a `float`. This would be beneficial for people that want exact durations (we still have this problem even without musical beats, e.g. for triplets), but would make the processing of values slower. I would make it another option in `note_array_from_part` to choose if we want fractions or floats.

In the case of anacrusis, this is not the wanted behavior. I will correct it in order to have the first note not marked as downbeat and the time elapsed calculated as if there was a "complete" measure. 
A similar problem will happen when we have "split measures" for example in the case of a repetition mark or a key change in the middle of a measure. Those are a bit harder to detect, especially if the score contains other problems. But I will still handle the standard cases.

sildater:
ad `use_musical_beat` in the part object: that's fine with me if there is a straightforward way to unset it. maybe as parameter in `set_musical_beat(self, use_musical_beat=False,mbeats_per_ts=None)`? and rename it to `set_beat_type`?

ad floats: I wouldn't change it, it's just not very appealing :) If people need exact durations they can work with `onset_div` and `duration_div`which are integer values relative to the part div, usually the smallest unit necessary to represent all note values of a part as integers.

neosatrapahereje:
Hi! 

@sildater: Regarding your first point, if I understand correctly, when `use_musical_beat` is set to  `True`, the  `onset_beat` and  `duration_beat` fields in the note array will use the new "musical" beats instead of the "notated" ones, so there is not an extra column added. 

Since using musical beats might be useful in general, I think it would be convenient to be able to use directly from the part objects, i.e., 

```python
part = partitura.load_musicxml("path/to/score.musicxml")
part.use_musical_beat = True
note_array = part.note_array
```
instead of 

```python
part = partitura.load_musicxml("path/to/score.musicxml")

note_array = partitura.utils.note_array_from_part(part, use_musical_beat=True)
```

The first one might be a one line longer, but then we don't need to call another function. I think that might be easier for people that just want to load a score and get the note array. In any case, adding the `use_musical_beat` flag to `Part` does not break the old behavior (since `self.use_musical_beat=False` by default).

neosatrapahereje:
Coming back to the `use_musical_beat` flag, another possibility could be to use the flag to the default beat map for the score, something like

```python
@property
def beat_map(self):
    if self.use_musical_beat:
        return self.musical_beat_map
    else:
        # i.e., rename the current beat_map to notated beat_map, or something else
        return self.notated_beat_map
```

In this case, we would need to be explicit in other methods (particularly in all utils that create note arrays).

An advantage of this approach would be that we would consistently use either musical beats or notated beats throughout all other properties of the `Part` object.

And if setting the flag to `True` is a bit cumbersome, we could have it as a method, perhaps something like

```python
def use_musical_beat(self):
    if not self._use_musical_beat:
        self._use_musical_beat = True
    else:
        # inform the user that musical beat was set all along
        print("Musical beats were being used all along!")

def use_notated_beat(self):
    if self._use_musical_beat:
        self._use_musical_beat = False
    else:
        print("Notated beats were being used all along!")
```

@fosfrancesco, @sildater What do you think?

sildater:
This is fine for me too:
```
part = partitura.load_musicxml("path/to/score.musicxml")
part.use_musical_beat = True
note_array = part.note_array
```
I'm not against a part property `use_musical_beat`. What I found confusing was that changing *to* musical beats is done using `part.set_musical_beats()` changing *back* to notated beats is done using `part.use_musical_beats = False` and `note_array_from_part` still has its own flag for it instead of querying the part property.

Another small thing that just came up is the `use_musical_beat` part property currently only affects the note_array. Going by its name I would assume it changes for instance the `part.beat_map` too. Ah I see you just wrote something about that. Yes, I agree, both proposals :)

fosfrancesco:
It seems that we agree! So I will make the changes and I will also delete the flag `use musical beats` from `note_array_from_part`, to not have a similar thing in 2 different places.

fosfrancesco:
Ok, I've done the modifications!

neosatrapahereje (approved):
This version looks good! I think the use of metrical or notated beats is a more intuitive.

PR #65, opened 2021-12-02
added unzipping for compressed musicxml files from musescore

sildater:
this change assumes a main musicxml file in the zipped container with the same file name as the .mxl file and the extension .xml. As far as I can tell this is the case for compressed musicxml files exported from musescore.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -7,6 +7,9 @@ import logging
 import numpy as np
 from lxml import etree
 
+import os
+import zipfile
+
 # lxml does XSD validation too but has problems with the MusicXML 3.1 XSD, so we use
 # the xmlschema package for validating MusicXML against the definition
 import xmlschema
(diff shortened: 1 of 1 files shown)


partitura/io/importmusicxml.py
```diff
@@ -187,6 +190,11 @@ def load_musicxml(xml, ensure_list=False, validate=False, force_note_ids=None):
         A list of either Part or PartGroup objects
 
     """
+    
+    if os.path.basename(xml).endswith(".mxl"):
```
neosatrapahereje, line 194: Perhaps it would be better to use `os.path.splitext` instead of `endswith`? (to make everything as "platform" agnostic as possible. 
Another possibility would be to use a `try` and `except` (since we cannot guarantee that the files would have the appropriate extension)

partitura/io/importmusicxml.py
```diff
@@ -187,6 +190,11 @@ def load_musicxml(xml, ensure_list=False, validate=False, force_note_ids=None):
...
+        with zipfile.ZipFile(xml) as zipped_xml:
+            xml = zipped_xml.open(os.path.basename(xml).split(".")[0]+".xml")
```
neosatrapahereje, line 196: And the same here (since "." can be part of the file name and not be necessarily the extension)

partitura/io/importmusicxml.py
sildater: good point!

neosatrapahereje (approved):

PR #67, opened 2021-12-07
added load_score method

neosatrapahereje:
A prototype of a simple method for importing a score in all supported formats.


neosatrapahereje:
@sildater and @fosfrancesco: as we discussed the other day, I added a simple general loader for all supported score formats (MEI and Kern have not yet been included), since I end up having to use a `load_score` method in many projects, and it would be a good convenience function.

Instead of checking the extension of the file, the current implementation uses `try`-`except` to go through the different supported formats, thus making the implementation robust against non-standard extensions (e.g., `.mid` and  `.MID` for standard MIDI files, `.musicxml` and `.xml` for MusicXML, etc.). In general, since you can just change the extension of a file without changing its contents, I think that relying on the extension might lead to some unexpected behavior.

While I think that  `try`-`except` might not be particularly elegant, it is a [pythonic solution](https://stackoverflow.com/a/1835844). The problem with the current implementation is that it would make error handling/debugging a little bit more obscure for the user (i.e., if something goes wrong, it would be hard to know why).

The question is, should we 
* leave is as it is (with `try`- `except`);
* Use the extensions (and add warnings). We can try to anticipate the possible extensions;
* Find a way to display possible errors (e.g., using [traceback](https://docs.python.org/3/library/traceback.html), which is built-in).

sildater:
I agree this is very useful! I also like the try except and the extension-agnostic approach. Besides catching the exceptions we could change the error message to something more generic that indicates other possible failures (wrong xml schema, file-like objects, musescore missing, ...?). Assuming that if users want to debug an import error they can use the format-specific import function.

Minor things:
- there is a typo in line 19 in the function definition "ementes"
- add tests
- all functions work with str, bytes or os.PathLike filenames but import_musicxml (in the uncompressed version) also takes _io.TextIOWrapper objects (``` with open(filename) as f: part = partitura.load_musicxml(f)```) or TemporaryFile objects. Maybe this should be documented
- add mei import

fosfrancesco (commented):
I can see the advantages. I try to avoid that, but it's mostly for personal taste since, as you said, this is actually suggested in python. Just for information, I've checked music21 and they do the selection based on the file extension.

However if a user for example tries to load an MEI that fails (probably it will be a common situation for a first period), it's a bit tricky to know what happened.
To solve make it easier I would add at the beginning an exception dictionary. Then for every exception:
`except Exception as e:
except_dict["MIDI"] = e`
Then at the end when you raise the Value error, you can print a nice text where you enumerate all the exception for every format that partitura tried to load.

Also adding `Exception as e` will avoid the catching of other things like keyboard interrupt.

Beside that I approve the modification!

sildater:
Looks good!

PR #69, opened 2022-01-10
Kern

melkisedeath:
# Kern import parser
- importing humdrum *kern files with kern parser object.
- treating kern as text and then char numpy
- possibility for parallelization in the future
- changed key fifths to inf
- added corresponding unit tests


tests/data/kern/Bach-BWV_316.krn
```diff
@@ -0,0 +1,116 @@
+!!!COM: Bach, Johann Sebastian
```
neosatrapahereje, line 1: @melkisedeath: Do we have the right to add this file to the repo? I assume their license allows that, but probably we need to add an acknowledgement somewhere in the README/Documentation...

melkisedeath: Hmm good question. There are a couple of royalty free Kern scores out there we could use instead or I could just take or write segments for the tests. Definitely, the tests still need some work but I think I'm gonna add tests as the use cases and exceptions occur. So for the moment just let it be and I will edit the tests and test files as we go.

PR #82, opened 2022-02-16
General Score Unfolding Overhaul

sildater:
# General Score Unfolding Overhaul

This PR contains the following new things in partitura.score.py:

* ```class Segment(TimedObject)```: a segment of a part between two time points that are relevant to the repetition structure
  - ```add_segments(part)```: a function adding segments to a part based on existing TimedObjects: Repeat, Ending, DaCapo, DalSegno, Segno, AlCoda, Coda, Fine
  - ```get_segments(part)```: returns a dict of segments indexed by segment ID
  - ```pretty_segments(part)```: creates a pretty string describing the segments of a part

* ```class Path```: a Path object encodes a sequence of segments and utility needed to extend that sequence, if it hasn't reached an end of the part.
  - ```unfold_paths(*args)```: recursively unfold and multiply a Path into all possible Paths and store these in a list
  - ```get_paths(part, *args)```: call unfold_paths with the starting segment Path and return the resulting paths as parametrized (for options, see below)
  - ```new_part_from_path(*args)```: create a new part from a Path
  - ```new_scorevariant_from_path(*args)```: create a new ScoreVariant from a Path

* Updated functions from which now use the new unfolding:
  - ```iter_unfolded_parts(*args)```: iterator over all unfolded parts
  - ```unfolded_part_maximal(*args)```: returns the longest unfolded part
  - ```unfolded_part_alignment(*args)```: returns the unfolded part that best fits the note IDs in an alignment
  - ```make_score_variants(*args)```: creates a list of ScoreVariant objects

## Why?

Theses changes solve limit cases of unfolding/segmentation:
- Volta brackets with >2 endings are supported
- Nested repeats are supported
- Other navigation marks (segno, coda, capo, fine) are supported

Additionally, the changes make the following possible:
* get all unfoldings with some specific constraints:
  - all!
  - the longest!
  - the longest without repeats after a jump mark (dal segno, al cado, da capo)
  - all without repeats after a jump mark (dal segno, al cado, da capo)
  - the shortest
* create any unfolding:
  - pick any segment(s) you like from the list of segments (```print(partitura.score.pretty_segments(part))``` for an overview) 
  - add them to a path ```p = Path(["C","K","&",...], segments)```
  - create a new part from this path ```new_part = new_part_from_path(p, part)```
* get informative strings about segments and paths:
  - as above: ```print(partitura.score.pretty_segments(part))```  
  - for paths use ```print(path)``` or ```print(path.pretty())```  

## How?

The Segments add information about the structure and the possible jumps from every segment to the part. 
The Path parses this tree of possible unfoldings (with constraints, if desired).
The ScoreVariant creates a variant of the score based on the list of of segment IDs stored (or possibly computed) in a Path.

## Other Bits and Fixes:

- ```load_musicxml```: correct parsing of location attribute of barline tag
- ```load_musicxml```: correct parsing of navigation markers (Da Capo, Al Segno, Segno, Al Coda, Coda, Fine)
- ```score.py```: TimedObject subclasses for navigation markers
- ```ScoreVariant.create_variant_part()```: Ignore the following score elements for the new part: Repeat, Ending, ToCoda, DaCapo, DalSegno, Segment, Page, System
- ```ScoreVariant.create_variant_part()```: Check for previous clefs before adding new clef element
- modified the unfolding unittest and added a new more complex test for unfolding with jumps and three Volta brackets
- added ```part.measure_number_map```here
- added test for ```part.measure_number_map```
- marked five functions as DEPRECATED (TODO: remove before merge)


partitura/io/musescore.py
```diff
@@ -153,7 +153,7 @@ def render_musescore(part, fmt, out_fn=None, dpi=90):
 
     """
     mscore_exec = find_musescore3()
-
+    print(mscore_exec)
```
neosatrapahereje, line 156: same here, do we need this print?

partitura/score.py
```diff
@@ -252,6 +252,45 @@ def measure_map(self):
...
+            LOGGER.warning("No measures found, assuming only one measure")
+            if self.first_point is None:
+                t0, tN = 0, 0
+            else:
+                t0 = self.first_point.t
+                tN = self.last_point.t
+
+            measures = np.array([(t0, tN, 1)])
+
+        return lambda x: measures[np.searchsorted(measures[:, 1], x, side="right"), 2]
```
neosatrapahereje, line 292: Maybe using interp1d instead of `lambda` and `searchsorted` could be faster? I don't know if this is the case, though, so you can disregard this comment ;)

partitura/display.py
```diff
@@ -62,9 +62,9 @@ def render(part, fmt="png", dpi=90, out_fn=None):
     """
 
     img_fn = render_musescore(part, fmt, out_fn, dpi)
-
+    print(img_fn)
```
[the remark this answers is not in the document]
sildater: not that I know. I didn't change this file, this is probably because of a mistake in a rebase.

partitura/io/musescore.py
sildater: not that I know. I didn't change this file, this is probably because of a mistake in a rebase.

partitura/display.py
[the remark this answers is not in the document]
sildater: reverted in a new commit.

partitura/io/musescore.py
sildater: reverted in a new commit.

partitura/score.py
```diff
@@ -2540,8 +2600,8 @@ def create_variant_part(self):
 
         return part
 
-
-def iter_unfolded_parts(part):
+# DEPRECATED
+def DEPRECATEDiter_unfolded_parts(part):
```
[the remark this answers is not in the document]
sildater: I'd completely delete them before merging, as they are replaced with equivalent functions of the same name and functionality, but with different internals. But renaming + warning/decorator is fine with me too.

partitura/score.py
sildater: This follows the structure of ```measure_map```.

melkisedeath: I think we should reconsider this lambda everywhere then. First, it is kind of counter-intuitive to call `part.func()` empty and then add another parenthesis for the arguments. Second, adding a lambda makes the new lambda function anonymous so makes it also hard for debugging. A solution could be to use `functools.partial` or just define all part functions to receive divs argument. Is there any particular reason I am not aware of for not doing this?

partitura/score.py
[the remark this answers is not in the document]
neosatrapahereje: ok, then I think it would be better to completely delete them before merging, since they could cause some confusion and will stop working altogether with newer versions.

partitura/score.py
fosfrancesco: I'm not familiar with functools.partial and interp1d. If you think it will be faster we can change that (also in measure_map). We just need to make sure that this can work both with one and multiple input values.

sildater: Both maps have property decorators so the double parentheses problem isn't there. I'm not familiar with partial functions either, but it's pretty straightforward to use interp1d and will work with single and array inputs, so I can change that.

fosfrancesco:
Good job, that was a lot of work. I found a possible problem by testing on this score 
image

It seems that `unfold_part_maximal(part)` is going twice over the 1st volta. So the resulting output pitches are:

`A F A F E E D E B B E E D E B B D`
while it should be
`A F A F E E D E B B E E D E D`

step to reproduce:
```part = partitura.load_musicxml("/content/repetition_test_xml.musicxml")
part_unf = partitura.score.unfold_part_maximal(part)
partitura.utils.ensure_notearray(part_unf,include_pitch_spelling = True)[["duration_beat","step"]]
```

[repetition_test_xml.zip](https://github.com/CPJKU/partitura/files/8087882/repetition_test_xml.zip)

sildater:
Good find! The problem in this file is that the first Volta bracket contains no starting tag in the musicxml, so ```load_musicxml```will discard the lone ending (see the warning when calling loading the file). After this the measures 3-4 are just a standard repeat, and the second Volta bracket has no effect on the repetitions structure. Should we add a guess for not started Volta brackets in ```load_musicxml```?

fosfrancesco:
I see, so this is MuseScore export problem (or maybe this is how it is supposed to be in musicxml for 1 measure volta? I don't know). I've tested it, and if the volta is more than 1 measure long, then Musescore will correctly produce also the `start`.  So probably a solution would be that if the start is missing, we set it to the beginning of the measure. And we can still print the warning, but maybe with a measure number attached if it's possible, so the error is more clear.

sildater:
I agree. And while we're at it we can adapt ```load_musicxml```to be able to deal with multiple repeats of the same bracket, like this example: https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/ending-element/

Issue #106, opened 2022-06-02
rest_array equivalent of note array property.


melkisedeath:
A complete rest array similar to note_array with an additional flag for "tying" rests.

```python
part.rest_array(include_time_signature=True, include_metrical_position=True, collapse=True)
```
The collapse flag joins consecutive rests on the same voice and keeps the id of the first one.

`part.note_array()` and `part.rest_array()` should be able to merge using `numpy.lib.recfunctions.merge_arrays()`

PR #108, opened 2022-06-05
add fast access properties to the part object

sildater:
added part properties for fast access (without having to use iter_all). The properties are called: measures, rests, repeats, dynamics, articulations, key_sigs, time_sigs. The follow the structure of notes and notes_tied, they return a list of objects.

The only unintuitive thing is that part.articulations returns part-level articulation directions only, text-based markings such as "staccato". Note-wise articulations such as a staccato dot (which are more common) are encoded as note attributes and not covered by this. To get these one has to use a note_array with articulation_feature from musicanalysis.note_features.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -283,14 +283,32 @@ def _parse_parts(document, part_dict):
                     "Single measure bracket is assumed")
 
         # complete unstarted repeats 
+        volta_repeats = list()
         for o in part.iter_all(score.Repeat, mode="ending"): 
             if o.start is None:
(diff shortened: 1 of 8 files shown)


sildater:
added another commit to this PR with changes that are not related to fast access, but fix some edge cases in repeat and volta processing.

melkisedeath:
This pull request allows accessing fast, properties of a part, such as rests, time signatures, key signatures, and others. A follow-up enhancement that will use these properties is `part.rest_array` #106 .

melkisedeath:
The extended-part-properties branch has unmerged commits causes Pull request to fail. Reopening Pull Request. @sildater

PR #110, opened 2022-06-06
Rest array

manoskary:
**rest_array** from part. Functions in `utils.music`:
- `rest_array_from_part`
- `rest_array_from_part_list`
- `rest_array_from_rest_list`
- `ensure_rest_array`

Also included property in _Part_ :
`Part.rest_array()`

`rest_array` has all fields of a `note_array`. For pitch and pitch spelling all values are set to 0.

### DIFF
--- a/partitura/utils/music.py
+++ b/partitura/utils/music.py
@@ -2232,7 +2232,7 @@ def rest_array_from_rest_list(
     Returns
     -------
     rest_array : structured array
-        A structured array containing note information. Pitch is set to 0.
+        A structured array containing rest information. Pitch is set to 0.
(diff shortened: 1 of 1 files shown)


melkisedeath:
> 

- Agreed about the docstrings. I reduced them but not completely deleted them. In case we want to extract documentation there should be there.
- The flags should stay there only to produce a rest array the same size as the note array. They are dummy attributes that take null values everywhere.
- Thanks for your remark about collapse and your provided test file. I did do some fixes but I believe it's not the optimal way to compute it. In any case, it should work now.

PR #112, opened 2022-06-07
Pianoroll from parts bug

sildater:
fixes bug in compute_pianorolls where the ordering of durations was different from the ordering of onsets.


fosfrancesco:
Nice catch! All seems to work for me. I'm merging

PR #113, opened 2022-06-08
Rest Features

melkisedeath:
This pull request adds some functionality concerning rests:
- Added Rest attribute is hidden for rests that should not be visible, i.e. not compiled in export.
- Created make_rest_features equivalent to make_note_features
- Included staff flag in `note_array_from_part`, `rest_array_from_part`, `ensure_notearray`, `ensure_rest_array`, `note_array_from_part_list`, `note_array_from_note_list`, `rest_array_from_part_list`, `rest_array_from_rest_list`. `None` staff value is initialized to 0.


sildater:
This PR adds hidden attribute for rests and rest_features equivalent to note_features.

PR #115, opened 2022-06-08
Note array to part

manoskary:
Function to create a part from a note array. Due to circular imports, I had to put it in `partitura.musicanalysis`. Please Review!


fosfrancesco:
I think it makes sense to have this in music analysis. This function is creating new material, possibly overlooking many other elements that may be in the part. This will be in future part of the MIDI to SCORE pipeline and probably could be moved in STRUTTURA. For the moment music analysis is a good place

melkisedeath:
This pull request should also include the incorporation of the `note_array_to_part` function into `midi_score_to_part`, and `match_to_part` when the latest version of match import is updated.

manoskary:
Checks are failing because in the current version of import midi with `note_array_to_part` addition the parts are not separated by track.

sildater:
Is this a final PR ready for review? Or are you working on it?

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,275 @@
...
+        note_ids = ["{}n{:4d}".format(part_id, i) for i in range(len(note_array))]
+        note_array = rfn.append_fields(note_array, "id", np.array(note_ids, dtype='<U256'))
+    elif assign_note_ids or np.all(note_array["id"] == note_array["id"][0]):
+        note_ids = ["{}n{:4d}".format(part_id, i) for i in range(len(note_array))]
+        note_array["id"] = np.array(note_ids)
+
+    dtypes = note_array.dtype.names
+    # check if note array contains time signatures
+    if not "ts_beats" in dtypes:
+        raise AttributeError("The note array does not contain a time signature.")
```
neosatrapahereje, line 169: Should the time signature information in the note array be required for this method? As we discussed last Friday, we could alternative just give an iterable with time signatures and times, and use that to create the time signatures instead (or if no time signature is added, assume a default time signature).

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,275 @@
...
+
+def create_part(
+        ticks,
+        note_array,
+        key_sigs,
+        part_id=None,
+        part_name=None,
+        sanitize=True,
+        anacrusis_divs=0
+):
```
neosatrapahereje, line 35: Maybe we could add some documentation to this function? (or point to `note_array_to_score` for info on the arguments. (Also, it would be great to add typing annotations in this module in general).

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,275 @@
...
+        warnings.warn("find and ensure tuplets", stacklevel=2)
+        # apply simplistic tuplet finding heuristic
+        score.find_tuplets(part)
+
+    warnings.warn("done create_part", stacklevel=2)
+    return part
+
+
+def note_array_to_score(note_array: Union[np.ndarray, list], part_id="", divs: int = None, assign_note_ids: bool = True,
+                       ensurelist: bool = False, estimate_key: bool = False, sanitize: bool = True, scorify=True) -> ScoreLike:
```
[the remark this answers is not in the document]
neosatrapahereje: It would be great to have arguments for specifying time and key signatures, in case this information is not provided in the note array, and otherwise, infer this information or set a default. (The most reliable thing would be to estimate the key signature and assign 4/4 as the default time signature).

neosatrapahereje (changes requested):
This is a very nice addition! There are a few things that might be worth changing:

1. Perhaps moving `note_array_to_score.py` to `utils`
2. Changing the requirement of time signatures in the input note array. It is good if this information is given, but otherwise have a dedicated argument for the time signature information.
3. Adding some documentation
4. Adding type annotations (we should try to be more strict on this)
5. Maybe add more tests to check edgecases (anacrusis, time signature changes, etc.)

partitura/musicanalysis/note_array_to_score.py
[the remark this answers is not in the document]
manoskary: What happens for now, is that if time signature is not available in the note array then it just fails but if we change the clause to something else we can enrich it later. I think it needs maybe a bit more thought on how to do this.

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,275 @@
...
+    onset_divs = list(map(lambda r: int(divs * r.numerator / r.denominator), onset_fractions))
+    min_onset_div = min(onset_divs)
+    if min_onset_div < 0:
+        onset_divs = list(map(lambda x: x - min_onset_div, onset_divs))
+    duration_divs = list(map(lambda r: int(divs * r.numerator / r.denominator), duration_fractions))
+    na_divs = np.array(list(zip(onset_divs, duration_divs)), dtype=[("onset_div", int), ("duration_div", int)])
+    return rfn.merge_arrays((note_array, na_divs), flatten=True, usemask=False)
+
+
+def create_part(
```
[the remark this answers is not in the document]
manoskary: Agree I'm going to add these things, straight away.

partitura/musicanalysis/note_array_to_score.py
manoskary: I thought it was a great idea at the time, but needed some brainstorming on deciding how to do it. Currently, you can input a note array in 3 different forms, we have to adapt everything to work with all three (i.e. beats, divs, or beats normalized by bar and global time). I left this open for future implementation.

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,275 @@
...
+        if n["duration_div"] > 0:
+            note = score.Note(
+                step=n["step"],
+                octave=n["octave"],
+                alter=n["alter"],
+                voice=int(n["voice"] or 0),
+                id=n["id"],
+                symbolic_duration=estimate_symbolic_duration(n["duration_div"], ticks)
+            )
+        else:
```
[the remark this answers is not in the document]
manoskary: Agree, very good comment I will add some tests on the input note array and also re-ordering.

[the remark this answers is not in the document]
manoskary: Actually, I don't understand the negative part. Grace notes should have zero durations. Only, non-zero durations are actual notes.

partitura/musicanalysis/note_array_to_score.py
neosatrapahereje: Ok, we should think how to properly include this in the next version.

partitura/musicanalysis/note_array_to_score.py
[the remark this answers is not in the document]
neosatrapahereje: That is true.

neosatrapahereje (approved):

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+import warnings
+import numpy as np
+from typing import Union
+import numpy.lib.recfunctions as rfn
+from fractions import Fraction
+import partitura.musicanalysis as analysis
+import partitura.score as score
+
+
+def create_divs_from_beats(note_array: np.ndarray):
```
sildater, line 12: could this helper go to utils?

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+        estimate key from note_array.
+    sanitize: bool (optional)
+        sanitize the part by adding measures, tying notes, and finding tuplets.
+    return_part: bool (optional)
+        Return a Partitura score object instead of a part.
+
+    Returns
+    -------
+    part : list or Part or PartGroup
+        Maybe should return score
```
sildater, line 209: update docstring

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+
+    if estimate_key or ('ks_fifths' not in dtypes and 'ks_mode' not in dtypes):
+        warnings.warn("key estimation", stacklevel=2)
+        k_name = analysis.estimate_key(note_array)
+        global_key_sigs = [[0, k_name]]
+    else:
+        if key_sigs is None and "ts_beats" in dtypes:
+            global_key_sigs = [[0, fifths_mode_to_key_name(note_array[0]["ks_fifths"], note_array[0]["ks_mode"])]]
+            for n in note_array:
+                if n["ts_beats"] != global_key_sigs[-1][1] or n["ts_beat_type"] != global_key_sigs[-1][2]:
```
sildater, line 311: 311 and 308 retrieve "ts_beats" while processing key signature. is that supposed to be like this? key sig and time sig seem unrelated.

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+        - onset_div or onset_beat
+        - duration_div or duration_beat
+        - pitch
+        - ts_beats
+        - ts_beat_type
+        - key_mode(optional)
+        - key_fifths(optional)
+        - id(required but can also be empty)
+    divs : int (optional)
+        Necessary Provided divs for midi import.
```
sildater, line 192: is divs optional or necessary?

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+    ks_end_times = np.r_[global_key_sigs[1:, 0], np.max(note_array["onset_div"]+note_array["duration_div"])]
+    global_key_sigs = np.column_stack((global_key_sigs, ks_end_times))
+
+    # compute quarter divs
+    if divs is None:
+        # find first note with nonzero duration (in case score starts with grace_note).
+        for idx, dur in enumerate(note_array["duration_beat"]):
+            if dur != 0:
+                break
+        divs = int((note_array[idx]["duration_div"] / note_array[idx]["duration_beat"])*(note_array[idx]["ts_beat_type"]/4))
```
sildater, line 326: if a note_array comes from a part set to `musical_beats`, this computation will give a wrong result

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+        anacrusis_divs = 0
+    else:
+        last_neg_beat = np.max(note_array[anacrusis_mask]["onset_beat"])
+        last_neg_divs = np.max(note_array[anacrusis_mask]["onset_div"])
+        beat_type = np.max(note_array[anacrusis_mask]["ts_beat_type"])
+        difference_from_zero = (0 - last_neg_beat) * divs * (4 / beat_type)
+        anacrusis_divs = int(last_neg_divs + difference_from_zero)
+
+    # Test again for negative divs
+    assert np.all(note_array["onset_div"] >= 0), "Negative divs found in note_array."
```
sildater, line 339: why is a second test required?

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+    note_array: np.ndarray
+        The note array from which the part will be created.
+    key_sigs: list
+        A list of key signatures. Each key signature is a tuple of the form (onset, key_name, offset).
+    part_id: str
+        The id of the part.
+    part_name: str
+    sanitize: bool
+        If True, then measures, tied-notes and triplets will be sanitized.
+    anacrusis_divs: int
```
sildater, line 64: this docstring is missing optionality indicators

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,358 @@
...
+        if renorm_onset_beat:
+            tmp = note_array[0]["onset_beat"]
+            tmp_ts = note_array[0]["ts_beats"]
+            for idx in range(1, len(note_array)):
+                if note_array[idx]["onset_beat"] < tmp:
+                    tmp_ts = tmp_ts + note_array[idx]["ts_beats"] if note_array[idx][
+                                                                         "onset_beat"] + tmp_ts < tmp else tmp_ts
+                    tmp = note_array[idx]["onset_beat"] + tmp_ts
+                    note_array[idx]["onset_beat"] = tmp
+        note_array = create_divs_from_beats(note_array)
```
sildater, line 270: on a second look I don't understand this at all. what is the field "global_score_time"? is it used anywhere else in partitura? this also seems to be one of the only two places where the field `"ts_beats"` is required, the other being in connection with key sigs below, which I commented too. Maybe it's possible to get rid of this condition and estimate it, like in create_part?

partitura/musicanalysis/note_array_to_score.py
manoskary: Ok, that is slightly problematic, would you have a proposal to fix this?

partitura/musicanalysis/note_array_to_score.py
manoskary: Second test is to verify no negative onset divs are present, which means that the input was faulty

partitura/musicanalysis/note_array_to_score.py
manoskary: This was something that I implemented back when I started writing this, based on a couple of Basis features plus a normalization per bar, could also be useful with generation per bar models, and other staff. For now, I would say it is completely useless but I'm just including it for future use, in case we want to create a score from a generative algorithm.

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,359 @@
...
+    onset_fractions = [Fraction(float(ix)).limit_denominator(256) for ix in note_array["onset_beat"]]
+    divs = np.lcm.reduce(
+        [Fraction(float(ix)).limit_denominator(256).denominator for ix in np.unique(note_array["duration_beat"])])
+    onset_divs = list(map(lambda r: int(divs * r.numerator / r.denominator), onset_fractions))
+    min_onset_div = min(onset_divs)
+    if min_onset_div < 0:
+        onset_divs = list(map(lambda x: x - min_onset_div, onset_divs))
+    duration_divs = list(map(lambda r: int(divs * r.numerator / r.denominator), duration_fractions))
+    na_divs = np.array(list(zip(onset_divs, duration_divs)), dtype=[("onset_div", int), ("duration_div", int)])
+    return rfn.merge_arrays((note_array, na_divs), flatten=True, usemask=False)
```
sildater, line 36: It should be documented that this only works if there are no time signature changes that affect the ratio of beat/div.

partitura/musicanalysis/note_array_to_score.py
sildater: I vote against including that, as far as I can tell it's undocumented, untested, and unused.

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,359 @@
...
+
+    # Note id creation or re-assignment
+    if "id" not in note_array.dtype.names:
+        note_ids = ["{}n{:4d}".format(name_id, i) for i in range(len(note_array))]
+        note_array = rfn.append_fields(note_array, "id", np.array(note_ids, dtype='<U256'))
+    elif assign_note_ids or np.all(note_array["id"] == note_array["id"][0]):
+        note_ids = ["{}n{:4d}".format(name_id, i) for i in range(len(note_array))]
+        note_array["id"] = np.array(note_ids)
+
+    dtypes = note_array.dtype.names
```
sildater, line 237: here dtypes is a variable but the same info is used in lines 224, 226, 230

partitura/musicanalysis/note_array_to_score.py
sildater: should it then go on top with the other input checks?

partitura/musicanalysis/note_array_to_score.py
manoskary: Not, quite the same they might be updated with the ids.

partitura/musicanalysis/note_array_to_score.py
manoskary: I will comment out this part of the code, however, this is something that would be very interesting to me and I would like to include it at some point.

partitura/musicanalysis/note_array_to_score.py
sildater: I just document it.

neosatrapahereje (changes requested):
This version looks quite nice! There are a couple of comments regarding checking that the durations are not negative and when to sort the array (these are very minor things), but otherwise it looks quite usable and covers probably most of the common cases.

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,505 @@
...
+            global_time_sigs = np.column_stack((global_time_sigs, ts_end_times))
+        elif global_time_sigs.shape[1] == 4:
+            pass
+        else:
+            raise ValueError("time_sigs is given in a wrong format")
+
+        # make sure there is a time signature from the beginning
+        global_time_sigs[0, 0] = 0
+    
+    # Test Note array for negative durations
```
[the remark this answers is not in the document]
sildater: good point, I'll add this

partitura/musicanalysis/note_array_to_score.py
```diff
@@ -0,0 +1,505 @@
...
+
+    # Note id creation or re-assignment
+    if "id" not in dtypes:
+        note_ids = ["{}n{:4d}".format(name_id, i) for i in range(len(note_array))]
+        note_array = rfn.append_fields(note_array, "id", np.array(note_ids, dtype='<U256'))
+    elif assign_note_ids or np.all(note_array["id"] == note_array["id"][0]):
+        note_ids = ["{}n{:4d}".format(name_id, i) for i in range(len(note_array))]
+        note_array["id"] = np.array(note_ids)
+
+    # Order Lexicographically
```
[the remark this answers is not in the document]
sildater: good idea, I'll sort by pitch and onset, then add idx, as I guess is standard in partitura.

partitura/musicanalysis/note_array_to_score.py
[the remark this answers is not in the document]
fosfrancesco: better to transform these asserts to "if ... raise Exception"  blocks as suggested in the second part of Carlos comment

PR #126, opened 2022-07-27
change clef number to staff

fosfrancesco:
now score.Clef has an attribute "staff" instead of "number"


melkisedeath (approved):
Instances of `Clef.number` changed to `Clef.staff`

PR #147, opened 2022-09-26
Unify score loading methods

neosatrapahereje:
This pull request consists of two contributions:

1. All input file keyword arguments in all  loading methods (i.e., `load_score`, `load_performance`, `load_musicxml` etc.) have been renamed to `filename`; and 
2. Add deprecation utils so that ensure that code using old keyword arguments (e.g., `fn`, `score_fn`, `xml`, `kern_path`, etc.) still works (but will raise a warning asking users to change to the new convention)


sildater:
Nice work! This PR adresses argument consistency in loading functions and adds deprecation utils.

PR #153, opened 2022-10-04
Standardize the arguments and interface of exporting methods

neosatrapahereje:
This pull request consists of 2 main changes:

* Update all methods for exporting files to work with the new  `Score` and `Performance` objects
* unify all argument names to have a consistent interface (old parameter names will raise a warning, but the code will still work)


partitura/io/importparangonada.py
```diff
@@ -0,0 +1,209 @@
...
+            # Weird behavior trying to cast 0.0 as an integer
+            struct_array[c] = raw_array[1:, i].astype(float).astype(int)
+        else:
+            struct_array[c] = raw_array[1:, i].astype(dt[-1])
+
+    return struct_array
+
+
+@deprecated_alias(outfile="filename")
+def load_alignment_from_parangonada(filename) -> List[dict]:
```
sildater, line 76: this function should probably be named load_parangonada_alignment

partitura/io/exportparangonada.py
```diff
@@ -28,40 +48,78 @@ def alignment_dicts_to_array(alignment):
...
+    score_note_array: np.ndarray
+        The note array from the score. Only returned if `outdir` is None.
+    alignarray: np.ndarray
+    zalignarray: np.ndarray
+    featurearray: np.ndarray
     """
 
-    part = ensure_notearray(part)
-    ppart = ensure_notearray(ppart)
+    if isinstance(score_data, (Score, Iterable)):
```
sildater, line 110: ensure_note_array handles Score and Performance objects, why is here only the first item taken? Should this be documented?

partitura/io/exportmidi.py
```diff
@@ -71,16 +77,21 @@ def get_ppq(parts):
...
 def save_performance_midi(
-    performed_part, out, mpq=500000, ppq=480, default_velocity=64
-):
+    performance_data: PerformanceLike,
+    out: Optional[PathLike],
+    mpq: int = 500000,
+    ppq: int = 480,
+    default_velocity: int = 64,
+) -> Optional[MidiFile]:
     """Save a :class:`~partitura.performance.PerformedPart` instance as a
```
sildater, line 88: update doc, not only performedpart

partitura/performance.py
```diff
@@ -289,6 +335,68 @@ def __init__(
...
+                    (i, p.get("track", -1))
+                    for i, pp in enumerate(self)
+                    for p in pp.programs
+                ]
+            )
+        )
+
+        return n_tracks
+
+    def sanitize_track_numbers(self) -> None:
```
sildater, line 364: what is the use of this function? afaik partitura currently only loads into performances via load_performance_midi and load_match, in both cases only a single performedpart is in the performance. multiple performedpart objects are allowed for similarity with the score object and such that multiple midi files (e.g. many instruments) can be represented in a single object (akin to the score object for an orchestral piece). in this case the track numbers are not assumed to be unique per performedpart/midi file/instrument.

neosatrapahereje: Since `PerformedPart` objects can be created directly from note arrays, it would be possible to have multiple parts in the same performance (e.g., generating a bass part + a melody part). But if not  explicitly stated, the track information would be the same for each `PerformedPart` in the `Performance`. This method avoids would "sanitize" this issue, so that the track information in each of the `PerformedPart` objects does not overlap.

partitura/io/exportparangonada.py
neosatrapahereje: That is true. I was worried about the renaming of the note ids by  `note_array_from_part_list`, but then I remembered that it will not do anything if there is only a single part. Updated in commit [d637b99](https://github.com/CPJKU/partitura/commit/d637b995185186565a2bf54e35b38c51bde5f641)

partitura/performance.py
sildater: Ah so that's a helper for manual creation. Can we add a safety against too many tracks? only 16 are allowed I think.

neosatrapahereje: I think the limit is only in the channels, AFAIK, not on the tracks?

partitura/io/exportmidi.py
neosatrapahereje: Updated

partitura/performance.py
sildater: ah yes, that could be.

manoskary: Is this resolved?

partitura/io/__init__.py
```diff
@@ -106,19 +107,19 @@ def load_score(filename: PathLike, force_note_ids="keep") -> Score:
             print(exception)
 
         raise NotSupportedFormatError
-    
-    
+
+
 def lp(filename: PathLike) -> Part:
```
manoskary, line 112: should we use a better name for this? We also have the merge parts function in `score`.
Maybe `load_score_as_part`

partitura/performance.py
neosatrapahereje: Yes

manoskary (commented):
This PR, addresses a uniform repreentation of Scores and Performances throughout the package. New test files are added and i/o parangonada are heavily edited.

manoskary (approved):

Issue #159, opened 2022-10-10
Note array from note list with empty voice bug.


manoskary:
Note array from note list fails when voice field is empty. Fails because `voice.max()` is called.

Issue #158, opened 2022-10-10
Kern import fails


manoskary:
Kern files fail when notes have a symbolic duration of `0` or `breve` note duration.

PR #160, opened 2022-10-10
Kern import fix

manoskary:
This PR addresses issues #158 and #159 and other minor bug fixes.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -347,6 +347,26 @@ def _parse_parts(document, part_dict):
                     "Single measure bracket is assumed"
                 )
 
+        # Complete repeats without end.
+        volta_repeats = list()
+        for o in part.iter_all(score.Repeat, mode="starting"):
+            if o.end is None:
+                # if len(o.start.starting_objects[score.Repeat]) > 0:
+                #     starting = list(o.start.starting_objects[score.Repeat].keys())[0]
+                #     # if unstarted repeat from volta, continue for now
+                #     if len(starting.end.ending_objects[score.Repeat]) > 0:
+                #         # if repeat from volta, continue for now
+                #         volta_repeats.append(o)
+                #         continue
+
+                end_times = [r.start.t for r in part.iter_all(score.Repeat)] + [part._points[-1].t]
+                end_time_id = np.searchsorted(end_times, o.start.t+1)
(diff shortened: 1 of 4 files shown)


sildater (approved):

partitura/io/importmusicxml.py
```diff
@@ -347,6 +347,26 @@ def _parse_parts(document, part_dict):
...
+            if o.end is None:
+                # if len(o.start.starting_objects[score.Repeat]) > 0:
+                #     starting = list(o.start.starting_objects[score.Repeat].keys())[0]
+                #     # if unstarted repeat from volta, continue for now
+                #     if len(starting.end.ending_objects[score.Repeat]) > 0:
+                #         # if repeat from volta, continue for now
+                #         volta_repeats.append(o)
+                #         continue
+
+                end_times = [r.start.t for r in part.iter_all(score.Repeat)] + [part._points[-1].t]
```
sildater, line 362: r.start.t might not exist for all repeats (which was, for r.end.t in line 382, the error that made this change necessary), so maybe there should be an additional safety?

sildater (changes requested):

partitura/io/importmusicxml.py
manoskary: Although the `r.start.t` might have caused the initial error, the correct version of this score was only possible by adding the inverse loop of assuming repeat ends.  I don't know if those two loop cases might cause any other security threats.

Would you have any proposal for this??

manoskary: For context and for other possible reviewers/readers of this thread. The "bug" that was corrected concerns scores that have a repeat start at some position of the score but at end/Fine has no repeat signs, so when partitura musicxml parser searches how to close the repeat start with another `pt.score.Repeat` element it fails.

sildater: I added a safety in 005aba513427c7ea423a9cfc6a57c2ba55784e30

sildater: it's just so end_times doesn't loop over *all* repeats, but only well-positioned ones.

sildater: (it might work in a single list comprehension, I don't know, I'm editing in the browser.)

sildater (approved):

PR #163, opened 2022-10-19
add reference-based midi pitch to freq

sildater:
adds support for reference-based and tempered computation of frequencies from midi pitch information. improves over the previous version by 1) including arbitrary references 2) support of non-invertible temperament ratios tables (for arbitrary tables subtracting a maj2nd might be different from adding a min7th and subtracting an octave)

### DIFF
--- a/tests/test_synth.py
+++ b/tests/test_synth.py
@@ -6,6 +6,7 @@ import tempfile
 
 from partitura.utils.synth import (
     midi_pitch_to_natural_frequency,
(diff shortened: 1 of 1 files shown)


manoskary (approved):

manoskary:
Very nice addition. Do you think a unitest or adding an example to the docs would be useful for this?

sildater:
I thought it was overkill, but I can add one so the coverage doesn't go down :)

PR #164, opened 2022-10-19
div unit flag in load_performance_midi

sildater:
new flag to load performance midi files in midi div units. store their ppq values of midi files in performedparts. these changes don't affect other functionality (field names of note arrays are unchanged), but are sometimes useful for further preprocessing.


manoskary (approved):
PR to set custom quartet divs in Performance MIDI

Issue #166, opened 2022-10-21
Bug synthesizing scores with pickup measures


neosatrapahereje:
The method `sythesize` in `partitura.utils.synth` has a bug synthesizing score-like objects with pickup measures (since onsets in pickup measures have negative onset times in beats).

PR #167, opened 2022-10-21
Fix issue #166 (synthesizing scores with pickup measures)

neosatrapahereje:
This pull request contains the following changes

* It solves issue #166 (synthesizing scores with pickup measures) 
* It allows for using parameters for the tuning functions directly in the `synthesize` function
* It adds a new `performance_note_array_from_score_note_array` which allows for more flexibility converting score-like note arrays to performance-like note arrays (including adding extra features for non-constant tempo and velocity  in `performance_from_part`)


partitura/utils/music.py
```diff
@@ -2560,19 +2557,25 @@ def performance_from_part(part, bpm=100, velocity=64):
...
+        MIDI velocity. If a callable is given, the function is assumed to map
+        score time in beats to MIDI velocity. Default is 64.
 
     Returns
     -------
     ppart: PerformedPart
-
-    Potential extensions
-    --------------------
-    * allow for bpm to be a callable or an 2D array with columns (onset, bpm)
```
neosatrapahereje, line 2574: This pull request implements these extensions (now as part of the `performance_notearray_from_score_notearray` method below)

partitura/io/exportaudio.py
```diff
@@ -39,15 +41,29 @@ def save_wav(
...
-    tuning: {"equal_temperament", "natural"}
+        If "linear" or "exp", the methods `lin_in_lin_out` and `exp_in_exp_out`
+        in `partitura.utils.synth` will be used. Otherwise this argument should
+        be a callable. See `lin_in_lin_out` for more details.
+    tuning: {"equal_temperament", "natural"} or callable
+        Method for tuning. If "equal temperament" will use equally spaced
+        semitones, while "natural" will use natural ratios within an octave.
+        Otherwise it uses a callable. See `midi_pitch_to_natural_frequency`
+        and `midi_pitch_to_frequency` for more info on the "equal_temperament"
+        and "natural" tuning.
```
sildater, line 54: natural temperament is based on the reference pitch of A, this choice is not clearly documented and gives wrong results for many pieces where the base note should be different. I see two ways of fixing this, 1) using the `midi_pitch_to_tempered_frequency` function instead or 2) clearly documenting that this renders a version based on  reference pitch A with interval ratios given in `NATURAL_INTERVAL_RATIOS`

sildater: I added a symmetric 5-limit temperament table `FIVE_LIMIT_INTERVAL_RATIOS`

Issue #171, opened 2022-10-31
Empty png rendering images with musescore


neosatrapahereje:
Rendering scores with MuseScore generates empty PNG files.

```
import partitura as pt
score = pt.load_score(pt.EXAMPLE_MUSICXML)
pt.render(score, fmt="png", out_fn="test.png")
```

Attached is a generated file and the expected output (computed directly from MuseScore's command line util)

![test](https://user-images.githubusercontent.com/6311456/198967717-6cf0a720-c7d4-4887-98dd-009e883d9079.png)

![test_expected](https://user-images.githubusercontent.com/6311456/198967797-27f7665e-408b-4064-859e-ca0105b3bd0f.png)

The issue (at least in MacOS with MuseScore 3.6 seems to be that MuseScore actually generates two png files: the first one is an empty page (which is what Partitura shows) and the second one is the actual score).

PR #173, opened 2022-10-31
Fix musescore render

neosatrapahereje:
This pull request fixes issue #171 

It contains the following changes

* Add a method `concatenate_images` for concatenating png files generated by MuseScore into a single image. This method has the advantage that it is able to display multi-page works instead of just showing the first page as the current `render_musescore` does. The trade off is that it requires [PIL](https://pillow.readthedocs.io/en/stable/). The Pillow library is standard library for image processing but is not built-in, so I would consider including it in the requirements. For now there is a workaround that just returns `None` if PIL is not installed.
* Deprecate old keyword arguments in `render_musescore` and `partitura.display` to be consistent with import/export methods (i.e., inputs are `ScoreLike` objects and not just parts).


manoskary (approved):
This changes for musescore rendoer of scores.
It includes optional dependency of PIL if already installed by user and implements a Unit Test.

manoskary:
What would be a nice extra is some documentation on the visualization of scores @neosatrapahereje .

PR #175, opened 2022-11-02
Pitch class piano roll

neosatrapahereje:
This pull request adds a method for computing pitch class piano rolls

* It adds method `compute_pitch_class_pianoroll` that computes pitch class features
* The note indices in the piano roll (using`return_idxs=True` in both `compute_pianoroll` and `compute_pitch_class_pianoroll`) includes now a fourth column with the original MIDI pitch (useful e.g., when using the piano rolls as part of note-level alignment pipelines).
* Fixes a typo in the documentation of `compute_pianoroll` (the docstring for time_unit incorrectly stated "second" as a valid time unit instead of "sec")
Adds a utility for generating random performances (`generate_random_performance_note_array`).


manoskary (approved):
A Pitch Class Piano roll representation and a function to randomly generate a Pianoroll.

Issue #185, opened 2022-11-17
`measure_map` error for Scores with just one measure


neosatrapahereje:
There is an issue with the `measure_map` method of `Part` objects when there is only one measure in the `Part`. This error is due to scipy's `interp1d` method requiring at least 2 elements in its inputs. The following code results in an error:
```python
import partitura as pt
# The example score contains just one measure
score = pt.load_score(pt.EXAMPLE_MUSICXML)
measures = score[0].measure_map(0)
```

Issue #184, opened 2022-11-17
Note array from Score does not include metrical information when setting `include_metrical_position=True`


neosatrapahereje:
Calling the `note_array` method from a Score instance using the `include_metrical_position` argument does not actually result in the inclusion of the metrical position features.

```python
import partitura as pt

score = pt.load_score(pt.EXAMPLE_MUSICXML)
na = score.note_array(include_metrical_position=True)

# This is failing
print("is_downbeat" in na.dtype.names)
```

PR #186, opened 2022-11-17
Fix issues #184  and #185

neosatrapahereje:
This pull request addresses the following issues:

* It closes issue #184 (missing metrical information when generating note arrays from `Score` instances)

* It closes issue #185. Since the issue with scipy's `interp1d` also affects other time maps, this pull request adds a drop-in replacement for Scipy's `interp1d` method in `partitura.utils.generic`. This utility has the same interface/arguments as inter1d, but it also allows for creating an interpolation with inputs of length 1. For example

```python
import numpy as np
int_fun = interp1d(x=np.array([1]), y=np.array([2]))
```

will result in `ValueError: x and y arrays must have at least 2 entries`. 

The new interpolation utility solves issues generating time maps for `Part` objects (e.g., `measure_map`) when there is only a relevant object (e.g., measures for `mesure_map`, key signatures for `key_signature_map`, etc.) defined in the `Part` instance.  Furthermore, the utility also allows for specifying the output dtype of the array, which is relevant for time maps where the output is expected to be an integer. This pull request also includes updates in the relevant time maps in `Part`, which now use the `intper1d` replacement from `partitura.utils.generic`.


neosatrapahereje:
The pull request includes code from a different branch, which should not have been included here... 😅 I will commit the code again from a different branch

PR #187, opened 2022-11-17
fix issues #184 and #185

neosatrapahereje:
This pull request addresses the following issues:

* It closes issue #184 (missing metrical information when generating note arrays from `Score` instances)

* It closes issue #185. Since the issue with scipy's `interp1d` also affects other time maps, this pull request adds a drop-in replacement for Scipy's `interp1d` method in `partitura.utils.generic`. This utility has the same interface/arguments as inter1d, but it also allows for creating an interpolation with inputs of length 1. For example

```python
import numpy as np
int_fun = interp1d(x=np.array([1]), y=np.array([2]))
```

will result in `ValueError: x and y arrays must have at least 2 entries`. 

The new interpolation utility solves issues generating time maps for `Part` objects (e.g., `measure_map`) when there is only a relevant object (e.g., measures for `mesure_map`, key signatures for `key_signature_map`, etc.) defined in the `Part` instance.  Furthermore, the utility also allows for specifying the output dtype of the array, which is relevant for time maps where the output is expected to be an integer. This pull request also includes updates in the relevant time maps in `Part`, which now use the `intper1d` replacement from `partitura.utils.generic`.

### DIFF
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -18,7 +18,7 @@ from numbers import Number
 from partitura.utils.music import MUSICAL_BEATS
 import warnings
 import numpy as np
-from scipy.interpolate import interp1d, PPoly
+from scipy.interpolate import PPoly
 from typing import Union, List, Optional, Iterator, Iterable as Itertype
 
 from partitura.utils import (
@@ -44,6 +44,8 @@ from partitura.utils import (
     _OrderedSet,
     update_note_ids_after_unfolding,
 )
+
+from partitura.utils.generic import interp1d
 
 
 class Part(object):
@@ -201,9 +203,6 @@ class Part(object):
 
             kss = np.array([(t0, fifths, mode), (tN, fifths, mode)])
 
-        elif len(kss) == 1:
-            # if there is only a single key signature
-            return lambda x: np.array([kss[0, 1], kss[0, 2]])
         elif kss[0, 0] > self.first_point.t:
             kss = np.vstack(((self.first_point.t, kss[0, 1], kss[0, 2]), kss))
 
@@ -259,12 +258,10 @@ class Part(object):
             kind="previous",
             axis=0,
             fill_value="extrapolate",
+            dtype=int,
         )
 
-        def int_interp1d(input):
-            return inter_function(input).astype(int)
-
-        return int_interp1d
+        return inter_function
 
     @property
     def measure_number_map(self):
@@ -314,13 +311,14 @@ class Part(object):
             measures = np.array([(t0, tN, 1)])
 
         inter_function = interp1d(
-            measures[:, 0], measures[:, 2], kind="previous", fill_value="extrapolate"
+            measures[:, 0],
+            measures[:, 2],
+            kind="previous",
+            fill_value="extrapolate",
+            dtype=int,
         )
 
-        def int_interp1d(input):
-            return inter_function(input).astype(int)
-
-        return int_interp1d
+        return inter_function
 
     @property
     def metrical_position_map(self):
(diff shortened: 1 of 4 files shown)


partitura/score.py
```diff
@@ -201,9 +203,6 @@ def key_signature_map(self):
 
             kss = np.array([(t0, fifths, mode), (tN, fifths, mode)])
 
-        elif len(kss) == 1:
```
neosatrapahereje, line 204: these lines were removed since this case is covered by the new `interp1d` utility in `partitura.utils.generic`.

partitura/score.py
```diff
@@ -259,12 +258,10 @@ def measure_map(self):
             kind="previous",
             axis=0,
             fill_value="extrapolate",
+            dtype=int,
         )
 
-        def int_interp1d(input):
```
neosatrapahereje, line 264: these lines were removed since this case is covered by the new `interp1d` utility in `partitura.utils.generic`.

partitura/score.py
```diff
@@ -314,13 +311,14 @@ def measure_number_map(self):
...
         inter_function = interp1d(
-            measures[:, 0], measures[:, 2], kind="previous", fill_value="extrapolate"
+            measures[:, 0],
+            measures[:, 2],
+            kind="previous",
+            fill_value="extrapolate",
+            dtype=int,
         )
 
-        def int_interp1d(input):
```
neosatrapahereje, line 320: these lines were removed since this case is covered by the new `interp1d` utility in `partitura.utils.generic`. (See the added `dtype=int` above)

partitura/score.py
```diff
@@ -346,12 +344,10 @@ def metrical_position_map(self):
                 axis=0,
                 kind="linear",
                 fill_value="extrapolate",
+                dtype=int,
             )
 
-            def zero_fun(input):
```
neosatrapahereje, line 351: these lines were removed since this case is covered by the new `interp1d` utility in `partitura.utils.generic`.

partitura/score.py
```diff
@@ -4586,12 +4583,16 @@ def merge_parts(parts, reassign="voice"):
     note_arrays = [part.note_array(include_staff=True) for part in parts]
     # find the maximum number of voices for each part (voice number start from 1)
     maximum_voices = [
-        max(note_array["voice"], default=0) if max(note_array["voice"], default=0) != 0 else 1
```
neosatrapahereje, line 4589: reformatted by black

partitura/score.py
```diff
@@ -4586,12 +4583,16 @@ def merge_parts(parts, reassign="voice"):
...
+        max(note_array["voice"], default=0)
+        if max(note_array["voice"], default=0) != 0
+        else 1
         for note_array in note_arrays
     ]
     # find the maximum number of staves for each part (staff number start from 0 but we force them to 1)
     maximum_staves = [
-        max(note_array["staff"], default=0) if max(note_array["staff"], default=0) != 0 else 1
```
neosatrapahereje, line 4594: reformatted by black

partitura/score.py
```diff
@@ -2495,7 +2491,7 @@ def microseconds_per_quarter(self):
 
         """
         return int(
-            np.round(60 * (10 ** 6 / to_quarter_tempo(self.unit or "q", self.bpm)))
```
neosatrapahereje, line 2498: reformatted by black

manoskary (approved):
Very good addition, it solves the `note_array` from score issue and also computes rich note arrays from parts with a single measure that was failing before. It includes two tests.

PR #189, opened 2022-11-21
Mei verovio

fosfrancesco:
The MEI parser is now more robust and able to ignore missing rests in measures.

The support for verovio was added. But verovio is still not a requirement for partitura. If it's installed partitura will use it to preprocess MEI, otherwise it will parse the original file


manoskary:
Are the new MEI files, i.e. `tests/data/mei/CRIM_Mass_0004_3.mei` supposed to be in the test files?

fosfrancesco:
Yes, this is an open problem so I left it there to be faster to work on it when the time will come.
I can delete it if you prefer but this is still not something people will download when they pip install the package, so I guessed I could leave it

manoskary:
It's just that this particular piece is wrong. It has an error on the encoding of whole note triads.

fosfrancesco:
That's right. I deleted it.

fosfrancesco:
That's right. I deleted it.

manoskary (approved):

PR #190, opened 2022-11-22
Midi tick

sildater:
adds new fields  ```onset_div``` and  ```duration_div``` to  ```PerformedPart.note_array()``` and the corresponding load functionality to ```load_performance_midi()```. The time unit is MIDI divs / ticks as loaded from the MIDI file directly.

### DIFF
--- a/partitura/io/importmidi.py
+++ b/partitura/io/importmidi.py
@@ -128,7 +128,7 @@ def load_performance_midi(
     for i, track in tracks:
 
         t = 0
-        tdiv = 0
+        ttick = 0
         
         sounding_notes = {}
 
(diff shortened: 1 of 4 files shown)


sildater:
so using the same field name as for part note arrays was a *bad* idea. I renamed it all to  ```onset_tick``` and  ```duration_tick```.

Tests are added too, the tempo change warning is disabled, and some documentation improvements are made for the pedal threshold setter.

manoskary (approved):
This PR creates two separate fields in note array, `onset_tick` and `duration_tick` it removes the conversion of onset division to tick as a replacement of the field and implements a relevant test for duration ticks.

PR #191, opened 2022-11-22
fix musescore loading and executable

sildater:
(for windows)


manoskary:
Do we have a unittest for that?
Then it would be great and would also solve the unixsystem problems.

sildater:
I don't really know how to do this, can we install a musescore executable on the CI machines? If that's possible, I can add tests.

manoskary:
I think it could be a good idea to add additional dependencies of our CI machines to also check our optional dependencies.

I think on Linux it should be as simple as adding a:
```
sudo apt-get install musescore
```
As a command line or the workflow yaml file.

neosatrapahereje (approved):

Issue #178, opened 2022-11-07
note ids should be strings starting with "n"


sildater:
whether loading from scores, matches, or performances, the note ids should be strings and starting with an "n".

Issue #89, opened 2022-03-08
Attribute Error in Match 3.1 when reading Time Signature


manoskary:
Time signature looks like `info(keySignature,[cn, major]).` and causes Error when `partitura.load_match(path, create_part=True)`.

Error log :
```shell
Traceback (most recent call last):
  File "/Users/Documents/2_UNI/03_JKU-PhD/CPJKU/CPCODE/code/mozart_pre.py", line 124, in <module>
    match_pp, match_a, match_sp = pt.load_match(
  File "/Users/Documents/2_UNI/03_JKU-PhD/CPJKU/CPCODE/partitura/partitura/io/importmatch.py", line 1262, in load_match
    spart = part_from_matchfile(mf, match_offset_duration_in_whole=True)
  File "/Users/Documents/2_UNI/03_JKU-PhD/CPJKU/CPCODE/partitura/partitura/io/importmatch.py", line 1581, in part_from_matchfile
    for (ks_beat_time, ks_bar, keys) in mf.key_signatures:
  File "/Users/Documents/2_UNI/03_JKU-PhD/CPJKU/CPCODE/partitura/partitura/io/importmatch.py", line 1154, in key_signatures
    step=ks_info.group("step1"),
AttributeError: 'NoneType' object has no attribute 'group'. Did you mean: 'time_signatures'?
```

melkisedeath:
To address the problem the lines 1142 in develop should be changed:
```python
# remove brackets and only take
# the first key signature
ks = ksl.Value.replace("[", "").replace("]", "").split(",")[0]
```
The split function removes the type of the Time signature when it is defined in match version **3.1**  as :
```shell
info(keySignature,[cn, minor]).
```

The split function exists to address multiple Time Signatures but in this case there is only one.
The error can not be fixed easily and be backward compatible with match files which contain multiple Time Signatures. @huispaty 

This split looks like a hack and needs to be addressed.

Issue #86, opened 2022-02-25
save_match relies on unreliable score information


sildater:
```matchfile_from_alignment()``` (called from ````partitura.save_match()```) relies on existence and uniqueness of measure numbers, which cannot be guaranteed. See line 122 in partitura/io/exportmatch.py

sildater:
Carlos, I assign you because of the match import/export rewrite. just as a heads up

PR #192, opened 2022-11-29
Support for Matchfile Version 1.0.0

neosatrapahereje:
This pull request contains a complete overhaul of the import and export capabilities for matchfiles according to the new specification for version 1.0.0. This pull request contains the following changes

* add the following files
    *  `matchfile_base.py`: Comple re-definition of the base (abstract) classes for match lines. The 
    * `matchfile_utils.py` contains utilities for parsing and formatting the attributes for the different match lines
    * `matchfile_v0.py` contains the definitions of the match lines in versions < 1.0.0 (e.g., 5.0).
    * `matchfile_v1.py` contains the definition of match lines in versions >= 1.0.0 (up to but not including a future 2.0.0)
* `importmatch.py` and `exportmatch.py` have been re-structured to use the new definitions. The `load_match` method supports loading all known match files, while `save_match` will just export versions starting from 1.0.0. Support for exporting older versions will only be added if there is enough interest.
* The attributes of the matchlines have been renamed to follow the specification published in the paper presented at the MEC2022 conference.
* Extensive tests for parsing matchlines, add test for exporting matchfiles (and updated example match file in tests to version 1.0.0)
* Adresses the following issues:
    *  #178:  An "n" is added at the beginning of note ids of performed notes if there is none.
    * #89: The new `MatchKeySignature` class can parse and generate the key signatures according to multiple formats (tested an all key signatures from the Magaloff, Zeilinger, Batik and Vienna 4x22)
    * #86: The `save_match` method does not make use of measure numbers from the part, but instead creates new numbers sequentially according to the starting time of the measure


partitura/score.py
```diff
@@ -4482,6 +4482,12 @@ def unfold_part_alignment(part, alignment):
     if len(best_idx) > 1:
         best_idx = best_idx[unfolded_part_length[best_idx].argmin()]
 
+    # append "-1" to alignment if the score_id's in alignment
```
manoskary, line 4485: I that new functionality/syntax? *append "-1" to alignment if the score_id's in alignment*

tests/test_match_import.py
```diff
@@ -133,8 +158,1836 @@ def test_load_match(self):
...
+            # wrong value (string instead of integer)
+            midiClockUnits_line = "info(midiClockUnits,wrong_value)."
+
+            mo = MatchInfoV1.from_matchline(midiClockUnits_line)
+            self.assertTrue(False)  # pragma: no cover
+        except ValueError:
+            # assert that the error was raised
+            self.assertTrue(True)
+
+        try:
```
manoskary, line 332: Is trying except clauses a valid practice for tests? What if it fails for a different `ValueError` and you accept the test.

manoskary (commented):
These changes address the New Match File format, with tests and other minor additions mainly focusing on the Match File format. Overall it's an excellent addition with: 
- support for past versions; 
- better formatting of the Match import into nice objects for every important attribute;
-  a lot of new tests, 
-  very nice unifying utils;
-  Midi Tick to seconds (extra addition)
-  testing for the export
-  Separate import files for matchlines 0 and 1

However I would like also to raise some concerns:
- This PR adds around 4500 new lines of code substantially expanding the package with functionality that mostly concerns supporting past match versions.
- Tests include try-except clauses which might not be the best practice;

Finally, I would request a second review from someone else.

partitura/io/importmatch.py
```diff
@@ -3,1061 +3,134 @@
...
-
-
-def interpret_field(data):
+def get_version(line: str) -> Version:
     """
-    Convert data to int, if not possible, to float, otherwise return
-    data itself.
+    Get version from the first line. Since the
+    first version of the format did not include this line,
+    we assume that the version is 0.1.0 if version is found.
```
sildater, line 100: if version is *not* found

partitura/io/importmatch.py
```diff
@@ -1439,55 +265,152 @@ def load_match(
...
+        offset = midi_ticks_to_seconds(first_note.Onset, mpq=mpq, ppq=ppq)
+    else:
+        offset = 0
+
+    notes = [
+        dict(
+            id=note.Id,
+            midi_pitch=note.MidiPitch,
+            note_on=midi_ticks_to_seconds(note.Onset, mpq, ppq) - offset,
+            note_off=midi_ticks_to_seconds(note.Offset, mpq, ppq) - offset,
```
sildater, line 373: performedpart notes now contain two extra (optional) fields: note_on_ticks, note_off_ticks. This could be added here as well.

sildater (approved):

PR #193, opened 2022-12-01
Release 1.2.0

sildater:
This PR merges the changes for version 1.2.0 to main.

New features
--------------------------------------
* Load and save alignments, performances, and scores stored in match files (.match) of all available versions
* Support for mei loading via verovio (if installed)
* PerformedPart notes store timing information in both ticks and seconds
* Support for pitch class piano rolls
* New MIDI time conversion functions

Bug fixes
--------------------------------------
* Fix render via musescore (if installed)
* Fix bug slowing down musicxml export
* Fix consecutive tie bug in kern import
* Gracefully handle slur mismatch in kern import
* Fix metrical position flag in note arrays
* Fix in MEI import and handling multiple staff groups inside the main staffGroup
* Fix measure map for single measure parts


partitura/utils/music.py
```diff
@@ -3,15 +3,27 @@
...
 import numpy as np
 from scipy.interpolate import interp1d
 from scipy.sparse import csc_matrix
-from typing import Union, Callable
+from typing import Union, Callable, Optional, TYPE_CHECKING
 from partitura.utils.generic import find_nearest, search, iter_current_next
 
+from partitura.utils.misc import deprecated_alias
+
+if TYPE_CHECKING:
```
manoskary, line 19: Hello, I raise the minor issue but the following lines are not very clear to me:

```python
if TYPE_CHECKING:
    # Import typing info for typing annotations.
    # For this to work we need to import annotations from __future__
    # Solution from
    # https://medium.com/quick-code/python-type-hinting-eliminating-importerror-due-to-circular-imports-265dfb0580f8
    from partitura.score import ScoreLike
    from partitura.performance import PerformanceLike, Performance
    ```
Why are these optional imports?

manoskary (approved):
This Release 1.2.0 seems ok, I only have a minor comment about typing.
There are a lot of new tests which is nice. The new match import is here, which has already been tested with ASAP.

partitura/utils/music.py
neosatrapahereje: These imports are necessary for typing annotations (since the type of many input variables in the utilities that are `ScoreLike` or `PerformanceLike`, which would create circular imports.

neosatrapahereje (approved):

manoskary (approved):
Minor corrections for solving `load_performance` bugs.

PR #199, opened 2022-12-19
Add new util: Slice PerformedPart by time

iv-pil:
This contains a new util function - slice_ppart_by_time, similar to slice_notearray_by_time. 
It takes as input a PerformedPart object, start and end times in seconds and returns a new instance containing all note, program and control information between those timepoints in the original object, without affecting the input. 
Optionally (by default) reindexes notes in returned instance to start from 'n0'.


partitura/utils/music.py
```diff
@@ -2871,6 +2871,116 @@ def get_matched_notes(spart_note_array, ppart_note_array, alignment):
...
+    """
+    from partitura.performance import PerformedPart
+    from copy import deepcopy
+
+    if not isinstance(ppart, PerformedPart):
+        raise ValueError("Input is not an instance of PerformedPart!")
+
+    assert(start_time < end_time), "Start time not less than end time!"
+
+    ppart_slice = deepcopy(ppart)
```
manoskary, line 2909: A deepcopy of a larger object is done here to be trimmed down afterward, the slice can be built from scratch.

partitura/utils/music.py
```diff
@@ -2871,6 +2871,116 @@ def get_matched_notes(spart_note_array, ppart_note_array, alignment):
...
+        if pr['time'] >= start_time and pr['time'] <= end_time:
+            new_pr = pr.copy()
+            new_pr['time'] -= start_time
+            if ppq:
+                new_pr['time_tick'] = int(2 * ppq * pr['time'])
+            programs_slice.append(new_pr)
+
+    notes_slice = []
+    note_id = 0
+    for note in ppart.notes:
```
manoskary, line 2938: maybe it is better here to loop through `performance.note_array()` this is running in numpy so it can be done faster. 
I would propose using numpy functions such as:
```
import numpy as np
import partitura as pt
idx = np.where((note_array["onset_sec"] >= slice_start) & ((note_array["onset_sec"]+note_array["duration_sec"] <= slice_end))
new_note_array = note_array[idx]
ppart_slice = pt.performance.PerformedPart.from_note_array(new_note_array)
```

manoskary (dismissed):
This slicing of the Performed part can be integrated fast with a few lines by slicing the note array with numpy. I propose a re-write of the function with numpy to be added to the next minor release as a new feature.

partitura/utils/music.py
iv-pil: Thank you for pointing this out, the deepcopy is indeed unnecessary. I will re-write this to start from a new class instance.

partitura/utils/music.py
iv-pil: That was my initial idea, but unfortunately this proposed method would not work due to the loss of information when converting to `performance.note_array()` as there is no distinction between 'note_off' and 'sound_off'. This creates a new performed part with notes list which contain different 'note_off" values than the original ones.
Therefore I would propose sacrificing efficiency in this case in order to obtain accuracy.

manoskary: Maybe it would be interesting to have an optional inclusion of such parameters to the note array then, i.e.
`performance.note_array(include_sound_off=True)`.  I propose to move this to a discussion so we can decide how to add functionality to the package that will be more inclusive for future potential uses.

Issue #206, opened 2023-02-20
bug in load_musicxml


anusfoil:
The `load_musicxml` function can only unzip the `.mxl` file when the zipped `.xml` file has the same name as the `.mxl` file. This works fine for more standardized musicxml data from ASAP and Vienna 4x22, but for a lot of public scores sourced from musescore community it's not the case. Often after unzip we find the contained score named `score.xml` and in this case will generate error. 

```
with zipfile.ZipFile(filename) as zipped_xml:
    xml = zipped_xml.open(
        os.path.splitext(os.path.basename(filename))[0] + '.xml'
    )
```
The following snippet is a simple fix:
```
with zipfile.ZipFile(filename) as zipped_xml:
    contained_xml_name = zipped_xml.namelist()[-1]
    xml = zipped_xml.open(
        contained_xml_name
    )
```

sildater:
Good catch! The fix looks good too, you can directly open a PR to the develop branch.

PR #208, opened 2023-02-22
Fixed load_musicxml with different name

anusfoil:
Issue #206


manoskary:
Thank you for addressing this issue.

CHANGES.md
```diff
@@ -2,7 +2,6 @@ Release Notes
 =============
 
 
-
```
manoskary, line 5: Unsure why this file was edited however it doesn't seem like an important change

partitura/io/importmusicxml.py
```diff
@@ -197,8 +197,9 @@ def load_musicxml(
     if isinstance(filename, str):
         if zipfile.is_zipfile(filename):
             with zipfile.ZipFile(filename) as zipped_xml:
+                contained_xml_name = zipped_xml.namelist()[-1]
                 xml = zipped_xml.open(
-                    os.path.splitext(os.path.basename(filename))[0] + ".xml"
+                    contained_xml_name
```
manoskary, line 202: Very simple and nice fix? Should we address this issue with a test or not?

manoskary (approved):

CHANGES.md
manoskary: This was fixed in the latest push.

PR #209, opened 2023-03-10
fixed bug in unfolding 

sildater:
Fixed an unfolding bug in some Beethoven Sonatas. Extended musicxml import support of "fine" encodings. Added part.segments property.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -733,6 +733,8 @@ def _handle_direction(e, position, part, ongoing):
 
         if get_value_from_attribute(sd, "fine", str) == "yes":
             part.add(score.Fine(), position)
+        if get_value_from_attribute(sd, "fine", str) == "1":
+            part.add(score.Fine(), position)
         if get_value_from_attribute(sd, "dacapo", str) == "yes":
             part.add(score.DaCapo(), position)
         if get_value_from_attribute(sd, "tocoda", str) == "coda":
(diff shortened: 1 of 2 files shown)


partitura/io/importmusicxml.py
```diff
@@ -733,6 +733,8 @@ def _handle_direction(e, position, part, ongoing):
 
         if get_value_from_attribute(sd, "fine", str) == "yes":
             part.add(score.Fine(), position)
+        if get_value_from_attribute(sd, "fine", str) == "1":
```
neosatrapahereje, line 736: Maybe we can combine the conditions for fine (i.e., lines 734 and 736) into a single condition? Something like

```python
if get_value_from_attribute(sd, "fine", str) in ("yes", "1"):
    part.add(score.Fine(), position)
```

And maybe, if "1" is a valid value for the attributes, maybe it could come as well for "dacapo" or similar repetition symbols? (I don't know if this is the case, so we probably should not try to anticipate something that never happens). But maybe we can add a comment there for future reference in case we ever find a score where `get_value_from_attribute(sd, "dacapo", str)` has the value "1".

neosatrapahereje (approved):
The commit looks good to me! I had only one minor comment. It would be good to add some tests for this, but this might require to create a complicated test MusicXML, so as long as this pull request does not break any other tests, I'm ok with it.

neosatrapahereje (approved):

Issue #223, opened 2023-04-11
Issue using `partitura.score.merge_parts` with elements without explicit staff info


neosatrapahereje:
There seems to be an issue merging parts by staff (i.e., using `reassign="staff"`) if the certain non-note elements do not have staff information.

This is a minimal example to reproduce the issue (see attached MusicXML file):

```python
import partitura as pt

score = pt.load_musicxml("28.xml")
part = pt.score.merge_parts(score.parts, reassign="staff")
```

Running this code results in the following error message:

```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-922f1ddd307d> in <module>
----> 1 part = pt.score.merge_parts(score.parts, reassign="staff")

~/Repos/partitura/partitura/score.py in merge_parts(parts, reassign)
   4705                 elif reassign == "staff":
   4706                     if isinstance(e, (GenericNote, Words, Direction)):
-> 4707                         e.staff = e.staff + sum(maximum_staves[:p_ind])
   4708                     elif isinstance(
   4709                         e, Clef
```

The solution should be fairly straightforward: simply assign staff number 1 to the elements that do not contain information, so that `e.staff` is always an integer.

[28.xml.zip](https://github.com/CPJKU/partitura/files/11197054/28.xml.zip)

PR #224, opened 2023-04-11
fix issue in #223

neosatrapahereje:
This is pull request closes issue #223. 

The pull request consists of a simple solution (as suggested in the issue) by simply ensuring that elements in different parts have a default staff value of 1.


manoskary (approved):
Nice and simple bug fix

PR #255, opened 2023-06-05
removes torch dependency

sildater:


neosatrapahereje (approved):
This pull request removes torch dependency from the package (only used  in the `notewise_to_onsetwise` and `onsetwise_to_notewise` methods).

neosatrapahereje:
This pull request removes torch as an additional dependency (since it was only used optionally in two functions, and removing the dependency does not change the behavior of the package).

Issue #133, opened 2022-09-02
Partial Measure Indexing


manoskary:
In `.musicxml` _partial_ measures meaning measure line that occur in between proper measure positions (proper = the ones that respect the time signature) get a prefix, i.e. "X1". Inferring the measures in partitura skips treating these measures correctly. A proposition would be to add a prefix and then the measure to which they preceed.

PR #256, opened 2023-06-05
Fix for #133 parsing and indexing for partial/irregular measures

huispaty:
Measure object now has 'number' and 'name' property:
- number : always int starting from 1, representing a running count independent measure regularity/ volta endings
- name : string, ID of a measure in a given musicxml score file, can be non-number in case of volta endings, irregular measures (i.e., pickup measures in the middle of the piece)


partitura/io/importmusicxml.py
```diff
@@ -456,13 +456,14 @@ def _parse_parts(document, part_dict):
         #     shift.applied = True
 
 
-def _handle_measure(measure_el, position, part, ongoing, doc_order):
+def _handle_measure(measure_el, position, part, ongoing, doc_order, measure_counter):
+    # print('_handle_measure') # NOTE del
```
neosatrapahereje, line 460: Also remove lines commented out.

partitura/io/importmusicxml.py
```diff
@@ -456,13 +456,14 @@ def _parse_parts(document, part_dict):
...
     """
```
neosatrapahereje, line 461: Maybe it would be worthwhile to document the function a little bit, since there is a new parameter (`measure_counter`). It would be nice to fully document all functions at some point 😉

partitura/score.py
```diff
@@ -2400,22 +2401,27 @@ class Measure(TimedObject):
...
 
-    def __init__(self, number=None):
+    def __init__(self, number=None, name=None):
         super().__init__()
         self.number = number
+        self.name = name
 
     def __str__(self):
-        return f"{super().__str__()} number={self.number}"
+        return f"{super().__str__():16s} number={self.number} name={self.name}"
```
neosatrapahereje, line 2424: Why is the addition of the spaces necessary (the `:16s`)? Is it because of a specific formatting required by `pretty_print`?

neosatrapahereje (changes requested):
This looks like a nice addition! I have a couple of comments regarding extra commented lines and a formatting of the `__str__` representation of the `Measure`.

Issue #214, opened 2023-03-15
Connect partitura to MIDITok or other tokenizers


manoskary:
Have a function that integrates MidiTok under the hood with:
- `tokens = to_tokens(part, token_type)`;
- `part = from_tokens(tokens, token_type)`

PR #263, opened 2023-06-07
Token export

fosfrancesco:
Allows to produce tokens from a partitura part with miditok.
(All limitations imposed by the specific tokenizers, e.g. only 4/4 are still to be taken into account since no error will be raised on partitura)

Corrects a bug on midi export with a pickup measure with the modality "pad_bar" which was not setting correctly the bar position.

Introduces a new modality for midi_export where the time signature is changed, to ensure correct beat positions for incomplete measures in the middle of the piece.


manoskary (approved):
## Summary

This PR integrates MidiTok functionality to partitura, with a `tokenize` function that transforms a scorelike object to a tokenized representations.

## Specifics

- closes #214 although it does not import score from tokens
- adds additional functionality for dealing with anacrusis in midi export: If `anacrusis_behavior=="time_sig_change"`, the time signature is changed to match the duration of the measure. It essentially sends a new MetaMessage with a time signature at the beginning of the pick-up measure.
- includes two new tests:
    1. To test incomplete measures
    2. Adds test for the Tokenizer comparing the exported tokens with the miditok tokens.

PR #265, opened 2023-06-09
change version in setup.py

### LINKED ISSUE #259: Partitura does not support Python <= 3.6
```shell
    from __future__ import annotations
    ^
SyntaxError: future feature annotations is not defined
python-BaseException
```

Linked googled [response](https://github.com/csuhan/ReDet/issues/14)

sildater:
is there anything else that needs to be changed?

Resolves #259


manoskary:
Should we also change in this PR  the release versioning stuff, or on the `develop -> main` PR would be best?

manoskary (approved):

Issue #264, opened 2023-06-09
tempo note_features breaks for file of vienna4x22


sildater:
```
perf, alignment, score = pt.load_match(filename=path, create_score=True)
feature_functions = ["loudness_direction_feature", "articulation_feature","tempo_direction_feature", "slur_feature"]
na =pt.compute_note_array(score, feature_functions=feature_functions)
```
gives error:
```
ValueError: field 'tempo_direction_feature.largo' occurs more than once
```
for match file https://github.com/CPJKU/vienna4x22/blob/master/match/Chopin_op10_no3_p01.match

PR #262, opened 2023-06-07
Closes #261 "none" mode for key signature.

manoskary:
Simple fix for `mode=='none'`.

Issue #261, opened 2023-06-07
Key Signature `none` in Match Export


manoskary:
The following issue occurs for empty Key signatures in match export.
```shell
...
File "/miniconda3/envs/asap/lib/python3.10/site-packages/partitura/io/matchfile_utils.py", line 689, in __str__
    ks = fifths_mode_to_key_name(self.fifths, self.mode)
  File "/miniconda3/envs/asap/lib/python3.10/site-packages/partitura/utils/music.py", line 736, in fifths_mode_to_key_name
    raise Exception("Unknown mode {}".format(mode))
Exception: Unknown mode none
```

Issue #257, opened 2023-06-05
Partitura does not load if music21 is not installed


neosatrapahereje:
There is an issue loading the package if music21 is not installed. This is because of the typing annotations

PR #268, opened 2023-06-09
Release 1.3.0

manoskary:
# New Release 1.3.0 (major)

This PR addresses release 1.3.0, it includes several bug fixes, code cleaning, documentation, and new functionality.

## New Features

- Enhanced Performance features in the same fashion as the note features;
- Fixed-size option for Note features. Use: `
- Create a score from a note array functionality. Call `partitura.musicanalysis.scorify(note_array)`;

## New Optional Features

- _If music21 is installed_ : Import music21 to Partitura by calling `partitura.load_music21(m21_score)`
- _If MidiTok is installed_ : Export Partitura Score to Tokens by calling `partitura.utils.music.tokenize(score_data, tokenizer)`

## Bug Fixes

- Fixed bug: #264 
- Fixed bug: #251 
- Fixed bug: #207
- Fixed bug: #162
- Fixed bug: #261
- Fixed bug: #262
- Fixed Issue: #256
- Addressed Issue: #133
- Fixed bug: #257
- Fixed bug: #248 
- Fixed bug: #223 

## Other Changes

- Minor Changes to the Documentation
- Addition of Docs link to the Github header
- Upgraded python version requirements to Python>= 3.7


manoskary:
Problem in uniformity:
Function call `partitura.musicanalysis.make_note_features` is not the same structure as `partitura.musicanalysis.compute_performance_features`

fosfrancesco (approved):

fosfrancesco (dismissed):

partitura/musicanalysis/note_features.py
huispaty, line 1: - In `make_note_features()` the `part` parameter is still named `part` but described as `score` in the function docstring.
- Proposal to rename `compute_note_array()` into `make_note_array()` in line with the other functions.
- In `loudness_direction_feature()` (and similarly, in the other `<name>_feature()` functions) it would be nice to also have the unknown descriptor displayed (i.e., `return f'{d.text}_unknown_constant'` in line 541)), or, alternatively (to not 'pollute' the note array), have it displayed when `force_size` is not `True` (add addtional `if/else` at the corresponding lines) to easier find/ fix potential mistakes in a score
- (minor remark) I find the parameter name `force_size` minimally misleading - what do you think about naming it `notnull`, `no_NaN` or something alike instead?

huispaty (approved):

partitura/musicanalysis/note_features.py
manoskary: Thanks for the feedback,

- I will change the name from part to score if it does not create import errors.
- The `compute_note_array` -> `make_note_array` might create some backward compatibility issues so probably it needs to be addressed appropriately later;
- I am not sure I completely understand the comment about the `<name>_feature`
- There was a related discussion about the `force_size` name, but I don't have any strong feeling about changing it. However, personally, I don't like the `notnull` or `no_NaN` flag alternatives.

manoskary: Renaming Part to score creates a Attribute error, so I leave it as it but I updated the documentation.

manoskary: Going back to the `<name>_unknown_feature` comment, if I understand better if we add the name to the string that is not contained in the constants then the size keeps growing for every new value. So then the force size does not hold.

manoskary: I leave @huispaty to resolve the conversation if the remarks are solved and then do the merge.

huispaty: Regarding remark no.3 (`loudness_direction_feature()` and similarly other `<feature_name>_feature()` functions (hope this way it is clearer): I just meant to include the unrecognizes string value together with `unknown constant` as the feature value in the note array, or (alternatively, to keep the note array "clean"), to return it also when the `force_size` argument is `False` (to make fixing potential score mistakes easier).

manoskary: I agree but I think the issue would persist, if we create those new names. The unknown constant can correspond to many different values, i.e. whatever is not already included in the constants. I would suggest to just merge and maybe we can address it for a later release.

huispaty (approved):

huispaty (approved):

fosfrancesco (approved):

Issue #272, opened 2023-06-15
several conditions are never reached in `feature_function_activation`


sildater:
`feature_function_activation` in `note_features.py` requires some reworking. the function starts with
```
if isinstance(
        direction, (score.DynamicLoudnessDirection, score.DynamicTempoDirection)
    )
```
and within this conditional block continues to check for instances of `score.ArticulationDirection` which isn't a subclass of those.
Minor styling: line 787 should also check a class.
There are open todos, likely related to issue #244

Issue #244, opened 2023-05-12
Unexpected drop in loudness_decr ramp


anusfoil:
When computing the dynamics direction feature `'loudness_direction_feature.loudness_decr'` for Chopin Op.10 No.3, there is an unexpected drop of the ramp value from 4 to 0 but there is no clear indication of the change on the score. To reproduce the problem: 

```
score = pt.load_musicxml('Chopin_op10_no3.musicxml')
na = pt.musicanalysis.note_features.compute_note_array(score, feature_functions=['loudness_direction_feature'])
na['loudness_direction_feature.loudness_decr']
```
Output is as follows (not sorted by time). From `n155` to `n156`, there is a sudden drop from value 4 to 0, which corresponds to no loudness decrease change at this point in score. Expected behavior: no change in the ramp value and 4 is kept towards the end of the piece. 

```
array([0.        , 0.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 0.        , 4.        ,
       4.        , 4.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 4.        , 4.        ,
       0.        , 4.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 0.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 4.        , 0.        ,
       4.        , 4.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 4.        , 4.        ,
       4.        , 0.        , 4.        , 4.        , 4.        ,
       4.        , 4.        , 4.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
       0.        , 0.        , 0.        , 0.        , 0.        ,
[...]

PR #274, opened 2023-06-15
update note feature extraction for dynamic features

sildater:
removes constant activation for dynamic score features. e.g. crescendi will no just start at 0, ramp up to 1 and drop back to zero at their end, not stay constant until a constant loudness direction is found.
This PR addresses the ramping behavior in #244 and also the problems mentioned in #272


manoskary (approved):
This PR looks good, it addresses issues about the dynamics ramp constantly increasing. Instead, the new approach informs about the beginning an end boundaries of a dynamics ramp such as crescendo or diminuendo

PR #276, opened 2023-06-15
addressed issue #275

### LINKED ISSUE #275: Assertion Error for computing performance features when tempo curve < 0
Hello, I got an `AssertionError` in the Zeilinger/Magallof subset which occurs when an element of the tempo curve is less than zero.
Reference on line 479 of `performance_codec.py`:
```python
assert((tempo_curve >= 0).all())
```

manoskary:
Fixes #275

### DIFF
--- a/partitura/musicanalysis/performance_codec.py
+++ b/partitura/musicanalysis/performance_codec.py
@@ -492,7 +492,11 @@ def tempo_by_average(
 
     tempo_curve = tempo_fun(input_onsets)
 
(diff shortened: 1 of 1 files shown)


sildater:
This assertion did its job and found a bug. The tempo curve is computed after monotonize_times is called in line 475, so it should be logically impossible for the tempo curve to be less than zero. Please check monotonize_times or the file instead of removing the assertion.

PR #284, opened 2023-06-16
gracefully deal with final notes that happen to be grace notes in `get_unique_seq`

sildater:
previous epsilon was flushed to zero, so the final duration was zero, leading to nan tempos.


sildater:
another shot at #275

manoskary (approved):
simple and elegant solution for grace note at the end of file edge case.

PR #296, opened 2023-07-28
Musescore4

fosfrancesco:
Added support for Musescore4.

In particular, partitura looks first for Musescore4, and if it is not present it throws a warning and it looks for Musescore3.

Moreover, a check was added in the Kern Parser, since when it was inputted with an XML file, it was not throwing any exception, but just returning a score with no parts. This was creating problems in the load_score function, because it was blocking the parser from reaching the Musescore condition.


partitura/io/importmei.py
```diff
@@ -68,6 +70,11 @@ def __init__(self, mei_path: PathLike) -> None:
         self.parts = (
             None  # parts get initialized in create_parts() and filled in fill_parts()
         )
+        # find the music tag inside the document
```
manoskary, line 73:
```suggestion
        # find the music tag inside the document
        music_el = self.document.findall(self._ns_name("music", all=True))
        if len(music_el) != 1:
            raise Exception("Only MEI with a single music element are supported")
        self.music_el = music_el[0]
```
What is a single music element, do you mean at least one note?

partitura/io/importmei.py
```diff
@@ -728,6 +744,53 @@ def _handle_mrest(self, mrest_el, position, voice, staff, part):
...
+        -------
+        position + duration : int
+            Next position on the timeline.
+        """
+        # find id
+        multirest_id = multirest_el.attrib[self._ns_name("id", XML_NAMESPACE)]
+        # find how many measures
+        n_measures = int(multirest_el.attrib["num"])
+        if n_measures > 1:
+            raise Exception(f"Multi-rests with more than 1 measure are not supported yet. Found one with {n_measures}.")
```
manoskary, line 777: This is function is described as handle for rests that spawn multiple measures but if the measures are more than one it raises an exemption, is that intended?

tests/test_load_score.py
```diff
@@ -37,9 +37,9 @@ def test_load_score(self):
             + MATCH_IMPORT_EXPORT_TESTFILES
             + EXAMPLE_FILES
         ):
-            self.load_score(fn)
+            self.check_return_type(fn)
 
-    def load_score(self, fn):
+    def check_return_type(self, fn):
         try:
```
manoskary, line 43: A try-except is not a nice test. I would prefer if we let it fail from the start, but this is not related to this PR.

tests/test_mei.py
```diff
@@ -33,7 +33,7 @@
 
 class TestImportMEI(unittest.TestCase):
     def test_main_part_group1(self):
-        parser = MeiParser(MEI_TESTFILES[5])
+        parser = MeiParser(MEI_TESTFILES[1])
```
manoskary, line 36: Indexing files is not a nice strategy for test as this might change in the future, I would say create a variable in tests init that describes the test or file.

tests/test_musescore.py
manoskary, line 1: If these tests are not performed then we should probably not include the file at all.

manoskary (changes requested):
This is a nice PR that addresses a lot of hidden issues and enhancements both in kern import, mei imprort and import via musescore.

- Solves a bug on kern returning empty score
- Enriches Mei import with most recent version and solves a couple of bugs
- Updates tests for mei
- Points some errors and vulnerabilities in tests.

Minor suggestions and comments are left for a final revision.

partitura/io/importmei.py
fosfrancesco: Music element refer to the XML element called music.

partitura/io/importmei.py
fosfrancesco: Yes, for the way the MEI and the parsing code are structured, handling multiple measures  in an error-robust way would require some major effort, which I will do for another release

tests/test_mei.py
fosfrancesco: I agree, but this should be done on another PR which change all tests

tests/test_load_score.py
fosfrancesco: I still corrected it to a better format

partitura/io/importmei.py
fosfrancesco: added <music> for better clarity

fosfrancesco:
I replied to the comments and changed things where it was possible!

tests/test_mei.py
manoskary: Yes, I will open a relevant issue for this so we can fix it for some next release.

manoskary (approved):
This PR addresses sume bug fixes in kern and mei import and adds support to the latest version of musescore. It is a nice addition to Partitura.

PR #302, opened 2023-08-03
added unfolding part minimal

manoskary:
Also changed `unfolding_part_minimal` and `unfolding part_maximal` to work with scores.

### DIFF
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -4595,7 +4595,7 @@ def unfold_part_maximal(score: ScoreLike, update_ids=True, ignore_leaps=True):
 
     Parameters
     ----------
-    score : :class:`Part`
+    score : ScoreLike
         The Part/Score to unfold.
     update_ids : bool (optional)
         Update note ids to reflect the repetitions. Note IDs will have
@@ -4609,8 +4609,8 @@ def unfold_part_maximal(score: ScoreLike, update_ids=True, ignore_leaps=True):
 
     Returns
     -------
-    unfolded_part : :class:`Part`
-        The unfolded Part
+    unfolded_part : ScoreLike
+        The unfolded Part/Score
 
     """
     if isinstance(score, Score):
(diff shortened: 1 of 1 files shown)


partitura/score.py
```diff
@@ -4588,15 +4588,58 @@ def iter_unfolded_parts(part, update_ids=True):
...
+        represent the first and second repetition of 'n132' in the
+        input `part`. Defaults to False.
+    ignore_leaps : bool (optional)
+        If ignored, repetitions after a leap are unfolded fully.
+        A leap is a used dal segno, da capo, or al coda marking.
+        Defaults to True.
+
+    Returns
+    -------
+    unfolded_part : :class:`Part`
```
sildater, line 4612: same here, for scores the return is also a score

partitura/score.py
```diff
@@ -4588,15 +4588,58 @@ def iter_unfolded_parts(part, update_ids=True):
...
+
+    paths = get_paths(
+        score, no_repeats=False, all_repeats=True, ignore_leap_info=ignore_leaps
+    )
+
+    unfolded_part = new_part_from_path(paths[0], score, update_ids=update_ids)
+    return unfolded_part
+
+
+def unfold_part_minimal(score: ScoreLike, update_ids=True, ignore_leaps=True):
```
sildater, line 4633: the update IDs and ignore leaps flags are superfluous I think. For ìgnore_leaps` this changes how repeats after e.g. a dal segno are treated, with `no_repeats` set to true this doesn't have an effect. same for `update_ids`: with no repeats, no IDs are used multiple times and no update is required.

partitura/score.py
```diff
@@ -4588,15 +4588,58 @@ def iter_unfolded_parts(part, update_ids=True):
...
+    """Return the "minimally" unfolded score/part, that is, a copy of the
+    part where all segments marked with repeat or volta signs are not included.
```
sildater, line 4635: the segments are included but not repeated, i.e. the unfolding takes the shortest path through the piece, but it doesn't omit segments.

partitura/score.py
```diff
@@ -4613,12 +4656,20 @@ def unfold_part_maximal(part, update_ids=True, ignore_leaps=True):
...
+        new_partlist = list()
+        for part in new_score.parts:
+            unfolded_part = unfold_part_minimal(part, update_ids=update_ids, ignore_leaps=ignore_leaps)
+            new_partlist.append(unfolded_part)
+        new_score.parts = new_partlist
+        return new_score
 
     paths = get_paths(
-        part, no_repeats=False, all_repeats=True, ignore_leap_info=ignore_leaps
+        score, no_repeats=True, all_repeats=True, ignore_leap_info=ignore_leaps
```
sildater, line 4669: `all_repeats` is also superfluous/overridden by `no_repeats`, maybe just remove it from the call

sildater (approved):

Issue #306, opened 2023-08-08
slow matchfile parsing due to duplicate lines check


huispaty:
Parsing match files takes longer than in previous versions due to the duplicate lines check in `importmatch.py`.  The attached test file (converted to .txt) takes > 3hrs to load.
[beethoven_op026_mv3.txt](https://github.com/CPJKU/partitura/files/12290282/beethoven_op026_mv3.txt)

manoskary:
I wrote the duplicate line check but that does not sound reasonable. Are you sure the issue is from the duplicate line checking?
I checked without the duplicate line validation but it still takes a long time (p.s. I didn't wait 3hrs though)

huispaty:
Yes I'm sure the issue stems from that part. Without this check, the file takes ~27secs to load (as it is quite large). With the checks, it takes >3.5hrs. I noticed this mainly because I was working on different branches, one of which had not yet been merged into `develop` (thus `main`) and had a version without this duplicate line check functionality.
My proposed solution looks like this:
```
    with open(filename) as f:
        raw_lines = f.read().splitlines()

    version = get_version(raw_lines[0])

    from_matchline_methods = FROM_MATCHLINE_METHODSV1
    if version < Version(1, 0, 0):
        from_matchline_methods = FROM_MATCHLINE_METHODSV0

    raw_lines = list(set(raw_lines))
    parsed_lines = [
        parse_matchline(line, from_matchline_methods, version) for line in raw_lines
    ]
    
    parsed_lines = [pl for pl in parsed_lines if pl is not None]

    mf = MatchFile(lines=parsed_lines)
```

Using this approach the same file takes ~25secs to load. Currently this is on my local branch only - it's not yet pushed as I would like to first address some raised issues that also relate to match file importing.

PR #307, opened 2023-08-10
Faster Matchifile Parsing and Checks.

manoskary:
This PR presents a solution for faster checking of empty and duplicate lines in matchfiles using numpy arrays and numpy vectorization.

- Addresses #306

### DIFF
--- a/partitura/io/importmatch.py
+++ b/partitura/io/importmatch.py
@@ -6,7 +6,7 @@ This module contains methods for parsing matchfiles
 import os
 from typing import Union, Tuple, Optional, Callable, List
 import warnings
-
+from functools import partial
 import numpy as np
 
 from partitura import score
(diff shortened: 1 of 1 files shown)


partitura/io/importmatch.py
huispaty, line 1: This changed version parses match files in the same amount of time as the version without any duplicate line checks (approx. 27 secs for the match file posted in the corresponding [issue](https://github.com/CPJKU/partitura/issues/306)). However, it adds quite a few additional lines and seemingly overcomplicates things... What about this alternative one-liner solution (also speeds up computation by a bit (~ 2 secs):

```
    with open(filename) as f:
        raw_lines = f.read().splitlines()

    version = get_version(raw_lines[0])

    from_matchline_methods = FROM_MATCHLINE_METHODSV1
    if version < Version(1, 0, 0):
        from_matchline_methods = FROM_MATCHLINE_METHODSV0

    raw_lines = list(set(raw_lines))
    parsed_lines = [
        parse_matchline(line, from_matchline_methods, version) for line in raw_lines
    ]
    
    parsed_lines = [pl for pl in parsed_lines if pl is not None]

    mf = MatchFile(lines=parsed_lines)
```

huispaty (dismissed):
The approach you've presented is efficient but could be simplified. If you're open to it, I would suggest considering the alternative one-liner I proposed earlier.

partitura/io/importmatch.py
manoskary: There are two issues your originally proposed code that I want to avoid with my addition:
1. Consecutive loops in pure python are slow (normally numpy functions should work faster maybe we can put several pieces to the test)
2. There are no warnings or indications about what is being discarded. The original version of the code (the slow one) had explicit warnings about repeated lines.

huispaty: 1. The two loops were part of the script previously (before the duplicate lines warning was introduced) but can be summarised into one easily.
2. I agree about the warnings part - though this is now reduced to merely indicating how many lines were dropped (instead of reporting which ones, as was done previously).

manoskary: [Here](https://stackoverflow.com/questions/46839277/series-unique-vs-list-of-set-performance) is a source comparing the `unique` with `list(set())`. Indeed for strings the overhead is not too much.

neosatrapahereje (changes requested):

partitura/io/importmatch.py
```diff
@@ -197,21 +197,31 @@ def load_matchfile(
...
+    np_lines = np.array(raw_lines, dtype=str)
+    # Remove empty lines
+    np_lines = np_lines[np_lines != ""]
+    len_after_empty_lines = len(np_lines)
+    if len_raw_lines != len_after_empty_lines:
+        warnings.warn(
+            f"Removed {len_raw_lines - len_after_empty_lines} empty lines from match file"
+        )
+    # Remove duplicate lines
+    _, idx = np.unique(np_lines, return_index=True)
```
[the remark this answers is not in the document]
manoskary: I think the lines are strings and therefore in numpy format they are read as python objects, therefore there would be easy sorting. Or maybe, I am missing something...

partitura/io/importmatch.py
```diff
@@ -197,21 +197,31 @@ def load_matchfile(
...
-            warnings.warn(f"Could not empty parse line: {line} ")
-            continue
-        parsed_lines.append(parsed_line)
-
+    len_raw_lines = len(raw_lines)
+    np_lines = np.array(raw_lines, dtype=str)
+    # Remove empty lines
+    np_lines = np_lines[np_lines != ""]
+    len_after_empty_lines = len(np_lines)
+    if len_raw_lines != len_after_empty_lines:
```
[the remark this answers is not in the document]
manoskary: Ok then, I will remove the warnings.

neosatrapahereje (changes requested):

neosatrapahereje (approved):

PR #310, opened 2023-08-23
mei measure number is now an integer, and. Measure.name is the string…

fosfrancesco:
Fixed a problem with Measure elements in MEI loading.
Now measure has 2 attributes:
- number: an integer, from 0, which increase at each measure
- name: the string how is encoded by the person writing the score (it may start from 1 or have strings inside)

### DIFF
--- a/partitura/io/importmei.py
+++ b/partitura/io/importmei.py
@@ -95,10 +95,11 @@ class MeiParser(object):
             raise Exception("Only MEI with a single score element are supported")
         sections_el = scores_el[0].findall(self._ns_name("section"))
         position = 0
+        measure_number = 1
         for section_el in sections_el:
             # insert in parts all elements except ties
-            position = self._handle_section(
-                section_el, list(score.iter_parts(self.parts)), position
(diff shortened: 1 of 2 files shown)


gregchapman-dev:
Can I recommend the name "index" instead of "number"?

fosfrancesco:
I think it is a good proposal to avoid confusion.

But this will need to be addressed in a separate pull request since we will need to update it also for all the other parsers, and some functions like ```measure_number_map```. I will open an issue for remembering and for discussion.

The goal of this pull request is to have the same behaviour as the musicxml parser.

huispaty (approved):
The suggested change works and corresponds to how we parse measures in musicxml. 
The `Measure` object has the `number` and `name` attributes. Currently, the `number` attribute represents a running count of measures (tags), irrespective of their regularity or volta endings - thus it starts at 1 and continuously increments.  However, we could also interpret `number` more akin to an index (as suggested [here](https://github.com/CPJKU/partitura/pull/310)) - and rename it to `index` accordingly. Let me know what you prefer.

PR #314, opened 2023-09-08
Load_score select the correct parser based on the file extension

fosfrancesco:
Load_score select the correct parser based on the file extension instead of try except

Solved a for load_score for match files, which was not returning anything

### DIFF
--- a/partitura/io/__init__.py
+++ b/partitura/io/__init__.py
@@ -35,7 +35,7 @@ def load_score(filename: PathLike, force_note_ids="keep") -> Score:
     """
     Load a score format supported by partitura. Currently the accepted formats
     are MusicXML, MIDI, Kern and MEI, plus all formats for which
-    MuseScore has support import-support (requires MuseScore 3).
+    MuseScore has support import-support (requires MuseScore 4 or 3).
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


partitura/io/__init__.py
```diff
@@ -54,20 +54,15 @@ def load_score(filename: PathLike, force_note_ids="keep") -> Score:
     scr: :class:`partitura.score.Score`
         A score instance.
     """
-    part = None
-
-    # Catch exceptions
-    exception_dictionary = dict()
-    # Load MusicXML
-    try:
+    extension = filename.split(".")[-1].lower()
```
manoskary, line 57: I would consider spliting the extension using the dedicated os function to avoid error with strings that have multiple `.` in them. I propose:
```python
os.path.splitext(filename)[-1].lower()
````

manoskary: Subsequently, all if clauses should contain the dot, i.e. `".xml", ".mxl", ...`

manoskary (changes requested):
This PR is a very useful addition, that should also avoid unnecessary import errors and speed up loading. However, a minor change is suggested.

fosfrancesco:
Changed as requested

partitura/io/__init__.py
fosfrancesco: agree!

PR #316, opened 2023-09-14
solution for empty dynamic directions.

manoskary:
This PR is a fix for a minor musicxml import bug that occurs when empty dynamics directions are encountered.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -842,10 +842,8 @@ def _handle_direction(e, position, part, ongoing):
             # dynamics items, so we loop:
             for child in direction_type:
                 # check if child has no children, in which case continue
-                if len(child) == 0:
-                    continue
                 # interpret as score.Direction, fall back to score.Words
-                dyn_el = next(iter(child))
+                dyn_el = next(iter(child), None)
                 if dyn_el is not None:
                     direction = DYN_DIRECTIONS.get(dyn_el.tag, score.Words)(
                         dyn_el.tag, staff=staff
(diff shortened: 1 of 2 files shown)


partitura/io/importmusicxml.py
```diff
@@ -841,6 +841,9 @@ def _handle_direction(e, position, part, ongoing):
             # first child of direction-type is dynamics, there may be subsequent
             # dynamics items, so we loop:
             for child in direction_type:
+                # check if child has no children, in which case continue
```
fosfrancesco, line 844: could you add a small test on when this was happening? Ideally a score with only the problematic measure inside. So we are sure that future partitura changes will not undo the fix

fosfrancesco (changes requested):

partitura/io/importmusicxml.py
fosfrancesco: Also, a probably nicer solution, next() accepts a default value. 
So if you do next(iter(), None) this will return None when the iterator is empty, and the None will be correctly handled by the next if statement

fosfrancesco: https://stackoverflow.com/questions/14250184/default-value-for-next-element-in-python-iterator-if-iterator-is-empty/14250244#14250244

manoskary: There is an issue with writing a test for this case since the musicxml export will not be the same as the import. This causes the test function  `test_import_export` to fail.

fosfrancesco: Very annoying. But then just don't add the piece to MUSICXML_IMPORT_EXPORT_TESTFILES. Instead create a new collection like MUSICXML_IMPORT_TESTFILES

manoskary: I added it implicitly to something else, so it can go through the coverage function. I don't think it should have its own file or function, it is too much.

fosfrancesco (approved):

fosfrancesco:
pull-request solving a bug in musicxml import when the direction tag contains no elements

Issue #315, opened 2023-09-14
Empty dynamics direction creates import error in musicxml


manoskary:
An empty dynamics direction creates an import error due to line a `next(iter(child))`

Issue #304, opened 2023-08-03
Issue when deepcopying a score


manoskary:
This problem occurs when `copy.deepcopy(score)` is called.
The issue results in a system error:
```shell
SystemError: <built-in method __deepcopy__ of numpy.ndarray object at 0x7f23580174b0> returned a result with an error set
```

To reproduce:

use this score : [here](https://github.com/musedata/humdrum-mozart-quartets/blob/master/kern/k080-01.krn) and call `copy.deepcopy(score)`.

sildater:
I don't really know where to start with this. I can reproduce the error and I also get for all randomly picked files from this repo, but not for other kern files. My guess is that either the parsing goes wrong for these files or there's a bug with numpy.

manoskary:
I will take a closer look at these files today once again and try to pinpoint which elements are causing this cloning error.

manoskary:
What I found is that the part structure is too deep and that is causing the recursion error, it can be fixed when the recursion limit is increased, like so : `sys.setrecursionlimit(10000)`

This is caused when the TimedObjects keep calling ending and starting objects on the timeline. Since any time objects can have many starting and ending objects which themselves will have other starting and ending objects then copy tries to pickle all these objects one by one but reaches a max recursion which could potentially be up to the number of TimedObjects in the score and therefore it fails. This is my best guess.

A potential fix is before calling copy to count the number of TimedObjects and set the recursion limit to that number.

Issue #297, opened 2023-07-28
Kern parsing mistake


manoskary:
Kern parser might parse XLM files and yield empty parts where it should return a parsing error.

Issue #293, opened 2023-07-15
Changing the 'note_off' property doesn't affect the 'duration_sec' property


Lelant:
I noticed something. Not sure if it is a bug. I have a PerformedPart. When I change the ```note_off``` property of a note, then the ```duration_sec``` property of the corresponding note_array() stays the same.

Here is an example of how I encountered this phenomenon:

```python
print(performedPart.notes[0])
print(performedPart.note_array()[0])

performedPart.notes[0]['note_off'] += 1.3

print(performedPart.notes[0])
print(performedPart.note_array()[0])
```

I print the values of the first note before and after changing the ```note_off``` value. I can see that this value has changed accordingly. The note_array() shows two types of duration, but not a note_off value. It shows the ```duration_sec``` and the ```duration_tick```. Now in my output I see that the value of ```duration_sec``` has **not** changed, but the value of ```duration_tick``` has changed as expected. 

Am I missing something or is this a bug or is this intentional? 

I noticed this because of using the Piano Roll. After changing the ```note_off``` value, the new created piano roll shows the same as before. So it seems the piano roll uses this ```duration_sec``` value, that hasn't changed. 

For now I found a workaround. For changing the length of a note I now use the note_array()'s property ```duration_sec``` instead of ```note_off```. And after the change I overwrite my notes of the PerformedPart with the changed note_array().
Like so:
```python
performanceNoteArray = performedPart.note_array()
performanceNoteArray[0]['duration_sec'] += 1.3
performedPart = pt.performance.PerformedPart.from_note_array(performanceNoteArray)
```

But I'm still curious if it should be possible to get a correct result in the piano roll just with changing ```note_off```

Thanks a lot,
Tim

Lelant:
I also noticed another related thing.

When I set the ```note_on``` and ```note_off``` values of notes of a performedPart by hand, and when I then try to compute the piano roll, I get the following error message: (coming from file partitura/utils/music.py, line 1316)

```python
raise ValueError("Note durations should be >= 0!")
```

But as I set the note_on and note_off values by myself, I know that they should result in positive durations for the notes. Here is how I did set the values:

```python
note_on_val = 0.0
note_off_val = 1.0
for note in performedPart.notes:
        note['note_on'] = note_on_val
        note['note_off'] = note_off_val
        note_on_val += 1.0
        note_off_val += 1.0
```

[...]

PR #320, opened 2023-09-22
PR for Release 1.4.0

huispaty:
New Features
------------
* new class for performed notes
* minimal unfolding for part
* updated Musescore parser for version 4
* `load_score` auto-selects parser based on file type
* new attributes for `Score` object for capturing meta information
* new score note attributes in matchfile export (`grace`, `voice_overlap`)
* new `tempo_indication` score property line in matchfile export

Bug Fixes
------------
* addresses #293
* solves #297 
* solves #304 
* solves #306
* solves #308
* solves #310
* solves #315

Other Changes
------------
* new unit test for cross-staff beaming for musicxml


huispaty:
Should we change the version to 1.3.2 or 1.4.0?

partitura/io/importkern.py
```diff
@@ -546,6 +546,9 @@ def parse_kern(kern_path: PathLike) -> np.ndarray:
     """
     with open(kern_path, encoding="cp437") as file:
         lines = file.read().splitlines()
+    if lines[0][0] == "<":
```
neosatrapahereje, line 549: Do we need this check? With the new `load_score` there should not be any issue. Alternatively, we could cover other cases and not just only MusicXML/MEI

tests/temp_test.py
```diff
@@ -0,0 +1,4 @@
+import partitura as pt
```
neosatrapahereje, line 1: We probably should delete this file, as it is not a proper test

tests/test_musescore.py
```diff
@@ -0,0 +1,31 @@
...
+from partitura.utils import compute_pianoroll
+from lxml import etree
+from xmlschema.names import XML_NAMESPACE
+from partitura.io import load_score, load_via_musescore
+import platform
+
+import numpy as np
+from pathlib import Path
+
+# This tests are commented, because it is not possible to run MuseScore on Github server
```
neosatrapahereje, line 22: Maybe we can just check whether MuseScore is installed and if so run the function? Otherwise, it might create potential conflicts with different versions of this script.

neosatrapahereje (changes requested):

partitura/io/__init__.py
```diff
@@ -76,44 +73,31 @@ def load_score(filename: PathLike, force_note_ids="keep") -> Score:
...
+    elif extension in [".kern", ".krn"]:
         return load_kern(
             filename=filename,
             force_note_ids=force_note_ids,
         )
-    except Exception as e:
-        exception_dictionary["Kern"] = e
-    # Load MuseScore
-    try:
+    elif extension in [".mscz", ".mscx", ".musescore", ".mscore", ".ms"]:
```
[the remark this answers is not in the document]
neosatrapahereje: So basically just add  `".kar", ".md", ".cap", ".capx", ".bww", ".mgu", ".sgu", ".ove", ".scw", ".ptb", ".gtp", ".gp3", ".gp4", ".gp5", ".gpx",  ".gp"` to the list 😉

partitura/io/importmidi.py
```diff
@@ -218,7 +217,7 @@ def load_performance_midi(
 
         # add note id to every note
         for k, note in enumerate(notes):
-            note["id"] = f"n{k}"
+            note["id"] = f"n{k+1}"
```
sildater, line 220: Is this necessary for something? It breaks every single alignment file in (n)ASAP and the link between performance midis and match files everywhere else?

partitura/io/musescore.py
```diff
@@ -36,38 +36,69 @@ class FileImportException(Exception):
...
+            )
+        else:
+            mscore_exec = find_musescore_version(version="")
+            if mscore_exec:
+                warnings.warn(
+                    "A unspecified version of MuseScore was found. Consider upgrading to musescore 4."
+                )
+            else:
+                raise MuseScoreNotFoundException()
+    if "DISPLAY" not in os.environ:
```
sildater, line 94: I end up at this exception on my laptop. (it has a display)

partitura/performance.py
```diff
@@ -280,6 +284,127 @@ def adjust_offsets_w_sustain(
...
+
+            self["duration_sec"] = self["note_off"] - value
+            self["note_on"] = value
+        elif key == "sound_off":
+            # Verify that the sound_off is after the note_off
+            if value <= self["note_off"]:
+                raise ValueError(f"sound_off must be after or equal to note_off")
+            self["sound_off"] = value
+        else:
+            self[key] = value
```
sildater, line 378: this function is never reached in my tests. if I rename is directly to `__setitem__` (is there a reason to use the standard setter in the init?), it breaks because it's full of recursions, e.g., line 378 `self[key] = value` calls `self.__setitem__` which should point to itself.

partitura/performance.py
```diff
@@ -280,6 +284,127 @@ def adjust_offsets_w_sustain(
...
+                f"Note on value provided is invalid, must be greater than or equal to 0"
+            )
+        if self["note_off"] < 0 or self["note_off"] < self["note_on"]:
+            raise ValueError(
+                f"Note off value provided is invalid, "
+                f"must be greater than or equal to 0 and greater or equal to note_on"
+            )
+        if self["velocity"] > 127 or self["velocity"] < 0:
+            raise ValueError(f"velocity must be between 0 and 127")
+
```
sildater, line 406: is this the solution you want? the same checks are made in the `__setitem__` , could they not be reused? the whole thing seems a bit hacky to me

tests/test_musescore.py
```diff
@@ -0,0 +1,33 @@
...
+  if find_musescore():
+    class TestImportMusescore(unittest.TestCase):
+        def test_epfl_scores(self):
+            score = load_via_musescore(MUSESCORE_TESTFILES[0])
+            self.assertTrue(len(score.parts) == 1)
+            # try the generic loading function
+            score = load_score(MUSESCORE_TESTFILES[0])
+            self.assertTrue(len(score.parts) == 1)
+except MuseScoreNotFoundException:
+    pass
```
sildater, line 33: related to the musescore checks above: this prevents the test from failing on my laptop, but musescore 4 is installed.

tests/test_tonal_tension.py
```diff
@@ -26,9 +26,9 @@ class TestTonalTension(unittest.TestCase):
     def test_prepare_notearray(self):
         target_note_array = np.array(
             [
-                (0.0, 2.0, 69, 64, "n0", "A", 0, 4, 0, -1),
-                (1.0, 1.0, 72, 64, "n1", "C", 0, 5, 0, -1),
-                (1.0, 1.0, 76, 64, "n2", "E", 0, 5, 0, -1),
+                (0.0, 2.0, 69, 64, "n1", "A", 0, 4, 0, -1),
+                (1.0, 1.0, 72, 64, "n2", "C", 0, 5, 0, -1),
+                (1.0, 1.0, 76, 64, "n3", "E", 0, 5, 0, -1),
```
sildater, line 31: my guess is this is related to the change in note ID attribution in performances. so the same as above, is this required?

sildater (changes requested):

sildater:
A general question, this PR links issue #293 which is not really addressed. The same behavior that was mentioned still persists. However, we probably also won't change it, so I just tried to give some details around it. How should we handle this? Should we keep this open? Or have some kind of FAQ section somewhere? I sometimes write these longer explanations on issues and would like them to stay around somewhere after we close the issue.

partitura/performance.py
manoskary: Yes, you are right. I wrote this but it doesn't make a lot of sense. What if we overwrite the `__setitem__` to include a `validate_values` with a key (to not check everything all the time) if it is provided:
```python
def __setitem__(self, key, value):
   self[key] = value
   self.validate_values(key)
```

partitura/performance.py
```diff
@@ -280,6 +284,172 @@ def adjust_offsets_w_sustain(
...
+    def _validate_note_on(self):
+        if self.pnote_dict["note_on"] < 0:
+            raise ValueError(
+                f"Note on value provided is invalid, must be greater than or equal to 0"
+            )
+
+    def _validate_note_off(self):
+        if self.pnote_dict.get("note_on", -1) < 0:
+            return
+        if (
```
sildater, line 417: maybe check for negative offsets also in this case?

partitura/performance.py
```diff
@@ -280,6 +284,172 @@ def adjust_offsets_w_sustain(
...
+            )
+
+    def _validate_note_on_tick(self):
+        if self.pnote_dict["note_on_tick"] < 0:
+            raise ValueError(
+                f"Note on tick value provided is invalid, must be greater than or equal to 0"
+            )
+
+    def _validate_note_off_tick(self):
+        if self.pnote_dict.get("note_on_tick", -1) < 0:
```
sildater, line 433: also here?

partitura/performance.py
```diff
@@ -280,6 +284,172 @@ def adjust_offsets_w_sustain(
...
+                self._validate_pitch()
+            elif key == "note_on":
+                self._validate_note_on()
+            elif key == "note_off":
+                self._validate_note_off()
+            elif key == "velocity":
+                self._validate_velocity()
+            elif key == "sound_off":
+                self._validate_sound_off()
+
```
sildater, line 401: add the tick time validators to the list.

partitura/performance.py
manoskary: I think I already do in the if clause.

partitura/performance.py
```diff
@@ -280,6 +284,172 @@ def adjust_offsets_w_sustain(
...
+        return self["note_on"] >= other["note_on"]
+
+    def __getitem__(self, key):
+        return self.pnote_dict.get(key, None)
+
+    def __setitem__(self, key, value):
+        if key not in self._accepted_keys:
+            raise KeyError(f"Key {key} not accepted for PerformedNote")
+        self.pnote_dict[key] = value
+        self._validate_values(key)
```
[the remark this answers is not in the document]
manoskary: Yes, this is indeed problematic. I pushed a proposed change on this it adds some additional clauses necessarily to avoid this issue.

partitura/io/musescore.py
manoskary: After internal discussion, the implemented solution is to check if the system is Linux and then check for display in `os.environ`. This is useful for most servers running Linux. The workaround for Windows or MacOS would be to use tkinter (a built-in library) but since such servers are not so common we will not implement it for now.

partitura/performance.py
sildater: yes, but only after this check. if the `note`_on`( or `note_on_tick`) is below 0, the second test is never reached and you can set an arbitrary offset.
another minor thing is that there is no upper limit for the onset setters, so there can be negative durations by setting higher onsets than offsets. 
this is only minor stuff, let me know if you want to fix it, it's fine to leave it as is too.

sildater (approved):

neosatrapahereje (approved):

Issue #267, opened 2023-06-09
match file version 1.0.0 voice parsing bug


sildater:
in the newest match file version voice information is stored as scoreattribute number with a prefix "v". this is not correctly parsed in load_match, where the voice information is cast to integers. The loading doesn't break but all voice information is lost.

sildater:
I'll add it here: the match files for chopin excerpts in 4x22 contain note IDs without dashes, e.g., "n1", the others contain repeats and are identified with dashes, e.g., "n1-1". It would probably be good to have this consistent and append a suffix for the ones without repeats.

manoskary:
Is this linked issue to any PR? Should it be scheduled for Release 1.3.2 ?

sildater:
@huispaty are you looking into this for the next release? 

some more things I just checked:
- voice information as part of the ScoreAttributesList is not documented in the match specification: https://cpjku.github.io/matchfile/
- I think the problem stems from `number_pattern = re.compile(r"\d+")` and 
```
            note_attributes["voice"] = next(
                (int(a) for a in note.ScoreAttributesList if number_pattern.match(a)),
                None,
            )
```
in `importmatch.py`
- I'm not sure whether this regex for voice number is used anywhere else 
- I'm not sure whether earlier match files encoded voice information beyond the "s" tag.
- my other comment about consistency in 4x22 might be moot because of the unfolding required foir Schubert and Mozart, just ignore it

PR #329, opened 2023-10-11
Voice parsing fix for bug #267

huispaty:
Fixes bug #267 for voice parsing from match files

### DIFF
--- a/partitura/io/importmatch.py
+++ b/partitura/io/importmatch.py
@@ -618,14 +618,15 @@ def part_from_matchfile(
         except (TypeError, ValueError):
             # no staff attribute, or staff attribute does not end with a number
             note_attributes["staff"] = None
-
-        if "v" in note.ScoreAttributesList[0]:
-            note_attributes["voice"] = note.ScoreAttributesList[0].split("v")[-1]
-        else:
-            note_attributes["voice"] = next(
-                (int(a) for a in note.ScoreAttributesList if number_pattern.match(a)),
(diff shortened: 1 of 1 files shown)


sildater:
Thanks for looking into this!

this PR has several problems:
- the implementation of voice parsing were voices are encoded as score attributes "vXYZ" is not working. Currently all match files of the newest version get all voice attributes set to 1 (the reason it's not `none` is the code in lines 747-750 in `importmatch.py`). Please check what strings the number_pattern regex is matching. If the logic is not clear, just ask. 
- it removes prior check for numbers (strings that only contain integers) in the scoreAttributeList. Are you sure this never happens in any match file? Unless you know for certain that the current implementation is unused or your replacement refactors the use case, please **do not remove or change used code and cases** from the library!  
- minor point but it'd still be nice: some unittest for different versions of match files would be great.

sildater (changes requested):
doesn't work yet. not automatically tested. removes potentially used functionality.

manoskary:
> doesn't work yet. not automatically tested. removes potentially used functionality.

The tests were not launched because the initial push was made for the main branch and then changed, an additional commit even an empty one, has to be made to launch the tests. I will do that.

sildater:
> > doesn't work yet. not automatically tested. removes potentially used functionality.
> 
> The tests were not launched 

Thanks! What I meant is the functionality in the PR is not addressed in a unittest.

Issue #286, opened 2023-06-20
Trill notes are not parsed as such from match v0


sildater:
In the example file `pianodata\match\beethoven_op002_no2_mv1.match` there are some trill markings. They do not show up in the alignment and in fact are probably not parsed correctly in line 199 of `importmatch.py`. A reason for the bug could be the regex for the line as defined in line 867 of `matchlines_v0.py`:
```
ornament_pattern: re.Pattern = re.compile(r"trill\((?P<Anchor>[^\)]*)\)-")
```
I think this is missing some regex for the performed note after the dash.

huispaty:
Do you have a testfile for this version? (The one in the [repo](https://gitlab.cp.jku.at/CP/pianodata/-/blob/master/match/beethoven_op002_no2_mv1.match) is of version 5.0)

sildater:
confusingly, v5.0 in match is v0 in partitura (v0 = v0.n.0 for n in 1-5), v1 is only used for v1.0.0

manoskary:
Is this linked issue to any PR? Should it be scheduled for Release 1.3.2 ?

huispaty:
No, it's neither linked to any existing PR nor scheduled for Release 1.3.2 (same for issue [#267](https://github.com/CPJKU/partitura/issues/267)). Both will be handled in the next release.

PR #332, opened 2023-10-23
fix pattern checking

sildater:
There was no checking for any specific patterns in deletions and insertions, just matching with respect the base pattern of notes and snotes. This meant that the classes:
- `MatchSnoteTrailingScore`
- `MatchSnoteNoPlayedNote`
were never used and always ended up as `MatchSnoteDeletion` and the classes:
- `MatchHammerBounceNote`
- `MatchTrailingPlayedNote`
- `MatchTrillNote`
were never used and always ended up as `MatchInsertionNote`. (which was mentioned in issue #286 )

The fix creates dummy instances in the base class methods to check for the correct patterns. This is a bit weird, but some of the classes only fix the patterns at instantiation, so I didn't know how else to get to the pattern. All above note classes are now correctly parsed and the parser further discards spurious lines such as `spurious_tag-note(606,[G,#],5,63037,63119,63119,67).` for not recognizing the pattern (this would previously have been an insertion).

### DIFF
--- a/partitura/io/matchfile_base.py
+++ b/partitura/io/matchfile_base.py
@@ -799,7 +799,20 @@ class BaseDeletionLine(MatchLine):
         matchline: str,
         snote_class: BaseSnoteLine,
         version: Version,
-    ) -> Dict:
+        pos: int = 0,
+    ) -> Dict:       
+        # this is very roundabout, but since this is a class method
+        # we can't use instance properties without a dummy instance of cls.
+        # and the note class pattern is only set when instantiated
+        # for some note classes (expecially v0)
+        dummy_snote = snote_class(version = version,anchor= 0,note_name="C",
+                            modifier="",octave=4,measure=0,beat=0, 
+                            offset=FractionalSymbolicDuration(1),duration=FractionalSymbolicDuration(1),
+                            onset_in_beats=0.0,offset_in_beats=0.0,score_attributes_list=[])
+        dummy_instance = cls(version = version, snote= dummy_snote)
+        match_pattern = dummy_instance.pattern.search(matchline, pos=pos)
+        if match_pattern is None:
+            raise MatchError("")
         snote = snote_class.from_matchline(matchline, version=version)
 
         kwargs = dict(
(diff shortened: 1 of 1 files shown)


partitura/io/matchfile_base.py
```diff
@@ -799,7 +799,20 @@ def prepare_kwargs_from_matchline(
...
         snote_class: BaseSnoteLine,
         version: Version,
-    ) -> Dict:
+        pos: int = 0,
+    ) -> Dict:       
+        # this is very roundabout, but since this is a class method
+        # we can't use instance properties without a dummy instance of cls.
+        # and the note class pattern is only set when instantiated
+        # for some note classes (expecially v0)
+        dummy_snote = snote_class(version = version,anchor= 0,note_name="C",
```
neosatrapahereje, line 808: Maybe it would be better to create the dummy instances outside of the method (i.e., as global variables)? Otherwise, calling this method will always create new instances of dummy_snote and dummy_instance.

partitura/io/matchfile_base.py
```diff
@@ -841,7 +854,24 @@ def prepare_kwargs_from_matchline(
         matchline: str,
         note_class: BaseNoteLine,
         version: Version,
+        pos: int = 0,
     ) -> Dict:
+        # this is very roundabout, but since this is a class method
+        # we can't use instance properties without a dummy instance of cls.
+        # and the note class pattern is only set when instantiated
+        # for some note classes (expecially v0)
+        if version >= Version(1, 0, 0):
```
neosatrapahereje, line 863: Same comment as above.

neosatrapahereje (changes requested):

partitura/io/matchfile_base.py
sildater: I can create the dummy snotes and notes outside but not the dummy instances for the class itself. so some instances are always created, yes

neosatrapahereje: Ok, in that case we can leave it as it is. If it is becomes too slow for large datasets (e.g., future extensions to (n)ASAP), maybe we should consider refactoring.

neosatrapahereje (approved):
Creating dummy instances for checking match errors would require a larger refactoring and potentially introduce other issues. This solution will stay for the time being, but we should consider refactoring some code for making the creating of matchlines more efficient. (This issue will be moved to a discussion).

PR #334, opened 2023-10-24
solution without dummies

sildater:
I found a solution for correct match line parsing that doesn't involve dummies, but additional identifying patterns that are independent of the match(s)note class and its version.


sildater:
also a minor change to discard unparseable lines, i.e., if `parse_matchlines` finds no matching pattern, it returns None, but we don't want these in the matchfile.

neosatrapahereje (approved):

Issue #337, opened 2023-11-19
Running tests locally without miditok


neosatrapahereje:
The new tokenizer functionalities require miditok and miditoolbox, which are not yet part of the requirements. Nevertheless, these packages are still used in `test_utils.py`, and result in an error when performing unittests locally.

manoskary:
Hello, I commented on the relevant PR #338 . This should be the intended behaviour. The online tests should run with no problem and check that all external dependencies work. Locally you would expect tests failing when you don't have a dependency with a message dependency X is missing.

PR #338, opened 2023-11-19
Check whether miditok is installed in tests

neosatrapahereje:
This pull request closes Issue #337.

It adds the following:

* It checks whether miditok is installed when running the tests
* Removes a duplicated test (`test_tokenize1`)

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -9,13 +9,24 @@ import numpy as np
 
 from partitura.utils import music
 from partitura.musicanalysis import performance_codec
-from tests import MATCH_IMPORT_EXPORT_TESTFILES, VOSA_TESTFILES, MOZART_VARIATION_FILES, TOKENIZER_TESTFILES
+from tests import (
+    MATCH_IMPORT_EXPORT_TESTFILES,
+    VOSA_TESTFILES,
+    MOZART_VARIATION_FILES,
+    TOKENIZER_TESTFILES,
+)
 
 from scipy.interpolate import interp1d as scinterp1d
 from partitura.utils.generic import interp1d as pinterp1d
 from partitura.utils.music import tokenize
-import miditok
-import miditoolkit
+
+try:
+    import miditok
+    import miditoolkit
+
+    HAS_MIDITOK = True
+except ImportError:
+    HAS_MIDITOK = False
 
 RNG = np.random.RandomState(1984)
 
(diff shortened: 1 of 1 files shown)


manoskary:
Hey, nice correction. I have only a minor comment on the if clause of the tests. I think that it doesn't make too much sense to make the tests conditional. For the automatic tests we should definitely verify that miditok is installed and the test is working instead of skipping, for the long term it is safer to not skip. For local tests it is ok just to let it fail maybe with a message that miditok is not installed. What do you think?

neosatrapahereje:
I actually think that it is not very elegant to let local tests fail. In the particular case of import errors (of packages imported at the top of the file), none of the tests in the file run, i.e.,  this is not reported by unittest as a test "failing" but as an *error*.  For the particular case of the miditok tests, none of the other tests  (e.g, the tests in`TestGetMatchedNotes`, `TestGetTimeMapsFromAlignment`, etc.) are actually run. This is, in my opinion, worse than checking whether a package is installed and conditionally run the tests.

manoskary:
I see the issue. What I would do is to check if miditok is installed and when is not assert true = false to make the tests fail with a message.
 To me the online tests are the most important so we can assure the robustness of the package in the long term when mistakes can happen and the workflow could change for any reason. I don't have a very strong opinion on that so if you want to do it the way you proposed is also fine by me.

sildater:
Hi! This conversation fell a bit silent, so let's try to find a solution. In my opinion this solution does exactly what we should do with optional libraries. Our required libraries can be used anywhere (they should be specified in the `requirements.txt` though), but for the others we have to create workarounds. Optional libraries include:
- music21 (currently just assumed in the test -> we should do that too)
- musescore (checked)
- miditok (this PR)
- verovio (checked)
- PIL (checked) 

I don't really know what the issue is with checking for an installed library, in the CI machine these libraries should be installed, no? well with the exception of musescore, but the optional python libraries should be. So the tests run and all is good.

manoskary:
Yes don't get me wrong, I completely agree. I was thinking about long term problems such as somebody editing the workflow without taking into consideration optional libraries and then a potential problem goes unnoticed. Apart from that edge case, then everything should work fine.

sildater:
true, I guess there is not that much room for error in the CI workflow and we have to be careful when changing something. Do you want to approve this PR or should I?

sildater (approved):

PR #341, opened 2024-01-10
add default tempo value for midi score export

sildater:


manoskary (approved):
Nice Simple solution.

PR #343, opened 2024-01-28
Track all MIDI messages for PerformedPart objects

neosatrapahereje:
This pull request allows PerformedPart objects to track all MIDI message types. In particular it allows for tracking the following information

* key signature
* time signature
* Meta MIDI messages

While this information is not necessarily reliable in performed MIDI files, it might be sometimes useful, in particular for generated music/performances.


manoskary (approved):
This PR adds Meta Messages about time tick, Key signatures and Time Signature in Midi performances for import and export. Its overall and nice and simple addition for completeness. It adds some typing and minor documentation

PR #347, opened 2024-01-31
Mei export v2

manoskary:
### MEI export (first simple version)
This method exports mei without full support for all elements as musicxml supports.
- Necessary to have a single part, or use `partitura.merge_parts` (maybe can this be changed)
- It supports clef, roman numerals (only for visualization), notes, grace notes, global elements (Time signature, key signature, cleff, staff)
- It supports clef changes
- It supports key and time signature changes only on the beginning of a measure
- It does not support beams unless a beam element is present.
- When note ids are available it assigns them otherwise it uses a unique `xml:id`


partitura/io/mei_export_v2.py
```diff
@@ -0,0 +1,383 @@
...
+            score_def_el.set(XMLNS_ID, "scoredef-" + self.elc_id())
+
+            # Add the scoreDef element at before the measure element starts
+            parent = measure_el.getparent()
+            parent.insert(parent.index(measure_el), score_def_el)
+            score_def_el.set('count', str(time_sig.beats))
+            score_def_el.set('unit', str(time_sig.beat_type))
+
+    def _handle_harmony(self, measure_el, start, end):
+        # For key signature changes, we add a new scoreDef element at the beginning of the measure
```
fosfrancesco, line 310: is this comment for another function?

partitura/io/mei_export_v2.py
```diff
@@ -0,0 +1,383 @@
...
+
+    Returns
+    -------
+    None or str
+        If no output file is specified using `out` the function returns the
+        MEI data as a string. Otherwise the function returns None.
+    """
+
+    if isinstance(score_data, spt.Score):
+        score_data = spt.merge_parts(score_data.parts)
```
fosfrancesco, line 347: I think it will not work here, because merge_part with default parameters will put all parts in a single staff. I would change this check and if the input contains more than one part, you raise an exception, that this is yet not supported. Something like
```
if isinstance(parts, Score):
        parts = parts.parts
    else:
        parts = list(iter_parts(parts)
if len(parts) > 1:
    raise ....
```

tests/test_mei.py
```diff
@@ -30,6 +30,28 @@
...
+            self.assertTrue(np.all(ina["voice"] == ena["voice"]))
+            self.assertTrue(np.all(ina["id"] == ena["id"]))
+
+    def test_export_with_harmony(self):
+        score_fn = os.path.join(MUSICXML_PATH, "test_harmony.musicxml")
+        import_score = load_musicxml(score_fn)
+        with TemporaryDirectory() as tmpdir:
+            tmp_mei = os.path.join(tmpdir, "test.mei")
+            save_mei(import_score, tmp_mei)
+
```
fosfrancesco, line 54: I approve this way of testing, but I would add at least one extra test with a different more standard score from our musicxml test scores. The EXAMPLE_MEI is somehow a very weird score

fosfrancesco (changes requested):
Good job for a first export! I left some comments

tests/test_mei.py
manoskary: I had some problems with making more tests in particular I got some conflicts from MEI import. More of them are related to renaming of the IDs during import. Although the note IDs are consistent in the exported MEI, the IDs during the MEI import change based on a namespace variable (`self.ns`). I will leave it to you if you want to make some changes in the import function.

partitura/io/mei_export_v2.py
manoskary: Addressed with commit [957df35](https://github.com/CPJKU/partitura/pull/347/commits/957df357accb8f1246e273afb8dfd2de592db4ef)

partitura/io/mei_export_v2.py
manoskary: Adressed with commit [273fd32](https://github.com/CPJKU/partitura/pull/347/commits/273fd325fcbc97a9a6fbbd3b74792e5efb3f6d42)

partitura/__init__.py
```diff
@@ -23,6 +23,7 @@
 from .io.importparangonada import load_parangonada_csv
 from .io.exportparangonada import save_parangonada_csv, save_csv_for_parangonada
 from .io.exportaudio import save_wav
+from .io.mei_export_v2 import save_mei
```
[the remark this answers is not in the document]
manoskary: Fixed with commit [ac52b06](https://github.com/CPJKU/partitura/pull/347/commits/ac52b063b784ccdc00c94d42a94e865c99a51a8e)

tests/test_mei.py
fosfrancesco: uhm...ok I will look into that!

fosfrancesco (approved):

PR #351, opened 2024-03-01
Mei export Updates

manoskary:
This PR solves some relatively important issues that were overlooked with MEI export.

- Added important support for dotted notes.
- Added Cadence Class and visualization in the export.
- Added support for Beams in MusicXML which will nicely visualize and export beams in MEI.
- Added support for barlines, title and fermata
- Added a function to infer beams
- Updated rest infilling.
- Added support for cross-staff beaming.

Thank you.


fosfrancesco (approved):
Adding functionalities for Cadence display, and correcting some problems in MEI export

PR #352, opened 2024-03-15
Estimate symbolic duration for Composite durations

manoskary:
This PR addresses a minor improvement on symbolic duration estimation for composite duration, i.e. tied notes or multiple rests.
We make this a standard on the rest infilling function in `score.py` and we generally make this an option. The PR includes some improved documentation.


partitura/io/exportmei.py
```diff
@@ -295,6 +296,22 @@ def _handle_tuplets(self, measure_el, start, end):
...
+                    "Tuplet start or end note is outside of the measure. Skipping tuplet element."
+                )
+                continue
+            if start_note.start.t > end_note.start.t:
+                warnings.warn(
+                    "Tuplet start note is after end note. Skipping tuplet element."
+                )
+                continue
+            # Skip if start and end notes are in different voices or staves
+            if start_note.voice != end_note.voice or start_note.staff != end_note.staff:
```
fosfrancesco, line 310: why should we skip if we are changing staff? This could be a staff crossing voice once we fix the modeling problem of partitura with these things.
I would allow it, unless it creates problems with the actual version

partitura/utils/music.py
```diff
@@ -319,6 +321,18 @@ class MIDITokenizer(object):
 # Standard tuning frequency of A4 in Hz
 A4 = 440.0
 
+COMPOSITE_DURS = np.array(
```
fosfrancesco, line 324: this should be a dictionary.
And we put a comment on top saying that this is a shortcut to make it work in common cases, but a more complex algorithm is necessary to handle general cases.

partitura/utils/music.py
```diff
@@ -915,10 +929,15 @@ def estimate_symbolic_duration(dur, div, eps=10**-3):
...
+    return_com_durations : bool, optional (default: False)
+        If True, return composite durations as well.
 
     Returns
     -------
-
+    out: Union[Union[Dict[str, Any], Tuple[Dict[str, Any]]], None]
+        Symbolic duration as a dictionary, or None if no matching
+        duration is found. When a composite duration is found, then it returns a tuple of symbolic durations.
+        The returned tuple should be tied notes.
```
fosfrancesco, line 940: "the returned tuple contains tied notes"

partitura/utils/music.py
```diff
@@ -941,14 +960,22 @@ def estimate_symbolic_duration(dur, div, eps=10**-3):
...
+            return copy.copy(SYM_COMPOSITE_DURS[j])
+        else:
+            # NOTE: Guess tuplets (Naive) it doesn't cover composite durations from tied notes.
+            type = SYM_DURS[i + 3]["type"]
+            normal_notes = 2
+            return {
+                "type": type,
+                "actual_notes": math.ceil(normal_notes / qdur),
+                "normal_notes": normal_notes,
+            }
```
fosfrancesco, line 978: does this function ever return None as specified in the function documentation?

fosfrancesco (changes requested):

partitura/utils/music.py
manoskary: Generally I agree, I just did it identical with the way it was done before for other durations for consistency.

partitura/utils/music.py
manoskary: Good catch, I have to update the documentation.

partitura/utils/music.py
manoskary: I am not sure I understand your comment. In the case of composite durations, the function returns a tuple of symbolic durations, these durations should correspond to many notes tied together. It does not return note objects.

partitura/io/exportmei.py
manoskary: This is true. However, it created some problems with the export. This is quite tricky. The issue comes from the Tuplet objects which should have tests implemented on them. Well, I can either comment out the staff and add a NOTE that this might create errors or add it as a flag on the function (i.e. cross_staff_in_tuplets=True).

partitura/utils/music.py
manoskary: Addressed with commit [61881a48](https://github.com/CPJKU/partitura/pull/352/commits/61881a482f6cc70fd2184f8a6546f22c0afd2172)

partitura/utils/music.py
manoskary: Addressed with commit [61881a48](https://github.com/CPJKU/partitura/pull/352/commits/61881a482f6cc70fd2184f8a6546f22c0afd2172)

fosfrancesco (approved):

PR #345, opened 2024-01-30
Dcml annotation parser

manoskary:
### Import for DCML corpora tsv files

**DONE**
- functional import with the possibility to include only notes, or notes, measures and harmony.
- Added optional dependency in pandas
- Some updates on score classes such as Cadences, Roman numerals and Phrases.
- Updated the estimate symbolic duration function in `music.py`.
- Move globals from `partitura.utils.music` to separate script called `partitura.utils.globals` to avoid circular import errors. (This will be done in a following PR after this one is merged to not break other components).
- Included infer root note, bass note from Roman Numeral and Chord Symbol strings. 
- Included validity checks for Roman Numeral Symbols and Cadence types.
- Updates and fixes on note transposition

**TODOs** (For new PRs)
- Remove pandas dependency: 
 This will be planned for later using `numpy.genfromtxt`. It can be done but the process is much more complicated because it transforms  many strings to byte strings and it will take some major restructuring. For now the code raises a warning that pandas is required. For the future, we replace with `numpy.genfromtxt` and replace the warning. A relevant Issue will be opened.
- Notebook and github discussion with documentation about RomanNumerals and Intervals and such

PR #353, opened 2024-03-28
Export wav using fluidsynth

neosatrapahereje:
This pull request adds the following functionality:

* Export a partitura object to an audio file using Fluidsynth. 
    * This functionality is available in [`pretty_midi`](https://github.com/craffel/pretty-midi/blob/main/pretty_midi/fluidsynth.py), where it only works for MIDI files. 
    * Such a method is useful for cross-modal tasks (e.g., transcription, audio-to-score alignment, etc.).
    * This implementation allows for the method to work on all partitura note-based objects (`Score`, `Performance`, `Part`, `PerformedPart`, note arrays), and thus for all formats supported by partitura. 
    * Fluidsynth is treated as an optional import. Instead of adding to the repository a default soundfont (as is the case in `pretty_midi`, which includes a very small one), the default soundfont is set to be downloaded from the [MuseScore soundfonts](https://musescore.org/en/handbook/3/soundfonts-and-sfz-files#list) (which are distributed with the MIT license).

* Fixes a few minor issues in `slice_ppart_by_time` adding previously missing MIDI messages and fixing the computation of onset and offset times in ticks.

### DIFF
--- a/partitura/utils/fluidsynth.py
+++ b/partitura/utils/fluidsynth.py
@@ -1,0 +1,265 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+"""
+This module contains methods for synthesizing score- or performance-like
+objects using fluidsynth. Fluidsynth is an optional dependency.
+
+"""
+
+import os
+from collections import defaultdict
+from typing import Callable, Optional, Union
+
+import numpy as np
+import partitura as pt
+
+try:
+    from fluidsynth import Synth
+
+    HAS_FLUIDSYNTH = True
+except ImportError:
+    Synth = None
+    HAS_FLUIDSYNTH = False
+
+from partitura.io.exportaudio import SAMPLE_RATE
+from partitura.performance import PerformanceLike
+from partitura.score import ScoreLike
+from partitura.utils.misc import PathLike, download_file
+from partitura.utils.music import (
+    ensure_notearray,
+    get_time_units_from_note_array,
+    performance_notearray_from_score_notearray,
+)
+from scipy.io import wavfile
+
+# MuseScore's soundfont distributed under the License.
+DEFAULT_SOUNDFONT_URL = "ftp://ftp.osuosl.org/pub/musescore/soundfont/MuseScore_General/MuseScore_General.sf2"
+
+DEFAULT_SOUNDFONT = os.path.join(pt.__path__, "assets", "MuseScore_General.sf2")
+
+if not os.path.exists(DEFAULT_SOUNDFONT) and HAS_FLUIDSYNTH:
+
+    download_file(
+        url=DEFAULT_SOUNDFONT_URL,
+        out=DEFAULT_SOUNDFONT,
+    )
+
+
+def synthesize_fluidsynth(
+    note_info: Union[ScoreLike, PerformanceLike, np.ndarray],
+    samplerate: int = SAMPLE_RATE,
+    soundfont: str = DEFAULT_SOUNDFONT,
+    bpm: Union[float, np.ndarray, Callable] = 60,
+) -> np.ndarray:
+
+    if not HAS_FLUIDSYNTH:
+        raise ImportError("Fluidsynth is not installed!")
+
+    if isinstance(note_info, pt.performance.Performance):
+        for ppart in note_info:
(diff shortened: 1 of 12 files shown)


tests/test_fluidsynth.py
```diff
@@ -0,0 +1,97 @@
...
+    SAMPLE_RATE,
+)
+
+from partitura import EXAMPLE_MUSICXML, load_score, load_performance_midi
+
+from partitura.io.exportaudio import save_wav_fluidsynth
+
+from tests import MOZART_VARIATION_FILES
+
+RNG = np.random.RandomState(1984)
```
manoskary, line 24: 1984 seems like a weird RandomState.... is that symbolic? Not a important comment

partitura/utils/misc.py
```diff
@@ -254,3 +256,27 @@ def concatenate_images(
...
+        The local file path where the downloaded file should be saved.
+
+    Notes
+    -----
+    This method was adapted from a Stack Overflow answer
+    (https://stackoverflow.com/a/15035466), and is distributed under the
+    CC BY-SA 4.0 license (https://creativecommons.org/licenses/by-sa/4.0/).
+    """
+    with urlopen(url) as in_stream, open(out, "wb") as out_file:
+        copyfileobj(in_stream, out_file)
```
manoskary, line 282: nice addition!

partitura/utils/fluidsynth.py
```diff
@@ -0,0 +1,326 @@
...
+
+    if not HAS_FLUIDSYNTH:
+        raise ImportError("Fluidsynth is not installed!")  # pragma: no cover
+
+    if isinstance(note_info, pt.performance.Performance):
+        for ppart in note_info:
+            ppart.sustain_pedal_threshold = 127
+
+    if isinstance(note_info, pt.performance.PerformedPart):
+        note_info.sustain_pedal_threshold = 127
```
manoskary, line 91: Should the sustain pedal threshold be an argument with a default value of 127? (just a suggestion maybe it is not applicable)

partitura/utils/fluidsynth.py
```diff
@@ -0,0 +1,326 @@
...
+
+# MuseScore's soundfont distributed under the MIT License.
+# https://ftp.osuosl.org/pub/musescore/soundfont/MuseScore_General/MuseScore_General_License.md
+DEFAULT_SOUNDFONT_URL = "ftp://ftp.osuosl.org/pub/musescore/soundfont/MuseScore_General/MuseScore_General.sf3"
+
+DEFAULT_SOUNDFONT = os.path.join(
+    pt.__path__[0],
+    "assets",
+    "MuseScore_General.sf3",
+)
```
manoskary, line 41: I think it would be nice if such arguments were added to a globals.py script. This is started on PR #345 it can also be done later.

manoskary (commented):
A very nice addition. This PR adds export wav with fluidsynth. It includes some useful utilities such as downloading a file.

partitura/utils/fluidsynth.py
neosatrapahereje: In this case, the reason for setting the pedal threshold to 127 is to have correct synthesis of note off and pedal messages. Otherwise, the synthesized signal would use "sound off", and create a very wet version of the performance (since both sustain pedal and sound off would produce a compounded effect).

partitura/utils/fluidsynth.py
neosatrapahereje: This seems like a good idea

tests/test_fluidsynth.py
neosatrapahereje: Well... I have used 1984 as my default random state for more than a decade (42 felt very obvious 😉). It is a reference to George Orwell's 1984 (since the book describes a totalitarian state, and I wanted my default rng to be everywhere) 🤓

partitura/utils/fluidsynth.py
manoskary: I think it will better to do this on a later PR when all other PRs are merged.

manoskary (approved):
Nice addition, a few comments and questions were already addressed and answered.

PR #361, opened 2024-07-19
add check for empty note array

sildater:
Added a ValueError in `compute_pianorolls` when passing empty note arrays.


sildater:
excuse me, the change is in `_make_pianoroll`.

manoskary (approved):
Valid extra assertion for empty note arrays in `_make_pianoroll`.

PR #366, opened 2024-08-01
Corrected Offsets for tied notes.

### LINKED ISSUE #365: Offsets for tied notes incorrect?
Python version 3.12.4
Partitura version 1.5.0

onset_div and onset_quarter for tied notes are reported seemingly incorrect

When loading the kern file ([test3.krn.txt](https://github.com/user-attachments/files/16457764/test3.krn.txt))  for following test fragment

Screenshot from 2024-08-01 16-04-11

and generating a note_array

```
import partitura as pt
import pandas as pd

score = pt.load_kern('test3.krn')

note_array = score.note_array(include_pitch_spelling=True, include_metrical_position=True)

pd.DataFrame(note_array)
```

this results in

Screenshot from 2024-08-01 16-10-18

Although there are 4 notes which (should) start at onset_div 0, only two do so, the ones being without a tie.

### LINKED ISSUE #367: Octave change not supported for clef (*clefGv2) - string compared to int for has_line
The [Humdrum Viewer documentation](https://doc.verovio.humdrum.org/humdrum/getting_started/clefs.html) states that "a vocal tenor clef is represented by *clefGv2, where the v means the music should be played an octave lower than the regular clef’s sounding pitches."

In [process_clef_line](https://github.com/CPJKU/partitura/blob/76f18c13bfaabb32632d52701d38f322d39e955f/partitura/io/importkern.py#L545)  we have 

https://github.com/CPJKU/partitura/blob/76f18c13bfaabb32632d52701d38f322d39e955f/partitura/io/importkern.py#L552-L565

On line 564, the match is extracted but that would be returned as a string, so the comparison on like 565 will always render False. 

Maybe by casting it to an int?

```
clef_line = int(has_line.group(0))
```

Included kern file fragment [test2.krn.txt](https://github.com/user-attachments/files/16468445/test2.krn.txt) that triggers this

manoskary:
Tied note mismatch during parsing of Kern scores. Corrected by omitting second loop for tie_prev.
Also corrected parsing of G clef with octave change.

Closes #365
Closes #367

### DIFF
--- a/partitura/io/importkern.py
+++ b/partitura/io/importkern.py
@@ -561,7 +561,7 @@ class SplineParser(object):
             else:
                 raise ValueError("Unrecognized clef line: {}".format(line))
         else:
-            clef_line = has_line.group(0)
+            clef_line = int(has_line.group(0))
         if octave_change and clef_line == 2 and clef == "G":
             octave = -1
         elif octave_change:
--- a/partitura/io/importkern.py
+++ b/partitura/io/importkern.py
@@ -435,12 +435,12 @@ class SplineParser(object):
             notes[self.tie_next], notes[np.roll(self.tie_next, -1)]
         ]:
             to_tie.tie_next = note
-            # note.tie_prev = to_tie
-        for note, to_tie in np.c_[
-            notes[self.tie_prev], notes[np.roll(self.tie_prev, 1)]
-        ]:
             note.tie_prev = to_tie
-            # to_tie.tie_next = note
+        # for note, to_tie in np.c_[
+        #     notes[self.tie_prev], notes[np.roll(self.tie_prev, 1)]
+        # ]:
+        #     note.tie_prev = to_tie
+        #     # to_tie.tie_next = note
         elements[note_mask] = notes
 
         # Find Slur indices, i.e. where spline cells contain "(" or ")"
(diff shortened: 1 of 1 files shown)


sildater:
simple fixes for two bugs during **kern import. Two minor questions:
- is there a reason to comment the faulty code instead of just deleting it?
- is there a reason to only accept treble Clefs on the second line for octave changes? It seems that **kern allows for [all options](https://doc.verovio.humdrum.org/humdrum/getting_started/clefs.html) and we could just parse them all by getting rid of the conditionals lines 565 - 571?

manoskary:
Thanks @sildater for your review and your comments!
To answer your questions:

> * is there a reason to comment the faulty code instead of just deleting it?

I was unsure if the second pass for tied notes could be helpful so I decided to keep it but you are right it is not useful so I deleted it and added a couple of comments.

> * is there a reason to only accept treble Clefs on the second line for octave changes? It seems that **kern allows for [all options](https://doc.verovio.humdrum.org/humdrum/getting_started/clefs.html) and we could just parse them all by getting rid of the conditionals lines 565 - 571?

In principle, a general parsing would be better but in practice, kern lines can take many forms since there are no strict formatting rules within kern files. In an attempt to parse as correctly as I could, I restricted the accepted solutions as much as possible. As the kern parser gets better maybe we can start loosening some of the constraints but for the moment I would suggest keeping it as is.

PR #370, opened 2024-08-28
Kern fixes (parsing editorials)

manoskary:
# Support for parsing Kern Files with Editorial Comments and Alternative Passages.

This PR addresses the parsing of Kern scores that include editorial comments such as alternative passages.
Such passages that could previously cause parsing errors can now be ignored and only the original passage or part is kept.

### Specific Comments on Code

- Two lines are commented out in `load_kern` which were inverting the part order, this has to be avoided in case the far right part contains only editorials. Instead, the part order is inversed when the partlist is passed to the score. Since this functionality changes, it should be monitored for potential errors.
- A Kern Object class is introduced for elements that do not correspond to partitura objects but describe the Kern parsing process within the format (e.g. editorials, or voice splitting/merging)

### DIFF
--- a/partitura/io/importkern.py
+++ b/partitura/io/importkern.py
@@ -527,9 +527,6 @@ class SplineParser(object):
     def process_tempo_line(self, line: str):
         return spt.Tempo(float(line))
 
-    def process_ossia_line(self):
-        return
-
     def process_fine(self):
         return spt.Fine()
 
--- a/partitura/io/importkern.py
+++ b/partitura/io/importkern.py
@@ -266,9 +266,6 @@ def load_kern(
     )
     # Get Splines
     splines = file[1:].T[note_parts]
-    # Inverse Order
-    # splines = splines[::-1]
-    # parsing_idxs = parsing_idxs[::-1]
     prev_staff = 1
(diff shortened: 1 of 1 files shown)


partitura/io/importkern.py
```diff
@@ -251,8 +267,8 @@ def load_kern(
     # Get Splines
     splines = file[1:].T[note_parts]
     # Inverse Order
```
fosfrancesco, line 269: can this be removed?

partitura/io/importkern.py
```diff
@@ -351,7 +368,7 @@ def load_kern(
     )
 
     doc_name = get_document_name(filename)
-    score = spt.Score(partlist=partlist, id=doc_name)
+    score = spt.Score(partlist=partlist[::-1], id=doc_name)
```
fosfrancesco, line 371: add a comment stating that we are inversing the order, and why we need to do that

fosfrancesco (changes requested):

partitura/io/importkern.py
manoskary: I think I had this somewhere.... thank I will add a comment

fosfrancesco (approved):

Issue #371, opened 2024-09-02
`estimate_symbolic_duration` breaks for some unknown durations


sildater:
loading `beethoven_op053_mv3.match` from the Zeilinger dataset throws an error in `estimate_symbolic_duration`. The inputs to reproduce without the file are `pt.utils.estimate_symbolic_duration(2496, 96)`. Looking into this, for most smaller fractions returns nonsensical values (like `{'type': 'long', 'actual_notes': 1, 'normal_notes': 2}`) if it doesn't hit a symbolic duration, for larger fractions the method fails. Test script:
```
not_working = 0
divs = 12
quarters = 32
for k in range(divs * 0, divs * quarters):
    try:
        a = pt.utils.estimate_symbolic_duration(k, divs)
        print(k, k/divs, a)
    except:
        print("not working:", k, k/divs)
        not_working +=1
print("percentage not working",not_working / (divs * quarters))
```

manoskary:
What should the expected duration or the intended behavior be?
I guess before partitura=1.5.0 when you had a failed case, the duration was assigned to `None`. 
In the example you mention, the resulting quarter duration of `2496 / 96`would be 26 quarter notes.
26 quarter notes should be a composite duration, i.e. a tied note. In the latest version of `estimate_symbolic_durations`, we can handle composite durations but our vocabulary is still limited. To guess composite duration you need to set  `estimate_symbolic_duration(return_com_durations=True)` then the output of the functions can be a list of dicts.

Currently, the `estimate_symbolic_duration` function has three clauses, actual possible durations, some kind of triplet duration, or composite durations. The choice was made so that the function does not return `NoneValue` and therefore is forced to choose a duration. Of course, it is very very far from being perfect with the idea being we keep adding new encountered durations as we find errors. Furthermore, the function is required to not fail so it can be able to guess arbitrary triplet durations, but as you've seen this does not work correctly in particular when composite durations are involved.

Another solution would be to allow the function to return `None` durations or empty dicts then whenever we cannot give a credible answer then we do not give any answer. I would be very open to having a deeper discussion on this topic to better structure the estimation so it could cover as many cases as possible and fit everybody's needs.

sildater:
Thanks for looking into this! I think the intended behavior should be to a) not raise exceptions and b) not give wrong values. `None` is good return value instead of the triplet condition which is the buggy part.

manoskary:
Ok, I think I have a solution, I will add you as the reviewer.

PR #372, opened 2024-09-05
New solution for estimate_symbolic_duration

manoskary:
Adressed issue #371

Introduced two new clauses that collapse the prediction to empty symbolic duration to avoid wrong values.

### DIFF
--- a/partitura/utils/music.py
+++ b/partitura/utils/music.py
@@ -762,8 +762,18 @@ def estimate_symbolic_duration(
         # 2. The duration is a composite duration
         # For composite duration. We can use the following approach:
         j = find_nearest(COMPOSITE_DURS, qdur)
-        if np.abs(qdur - COMPOSITE_DURS[j]) < eps and return_com_durations:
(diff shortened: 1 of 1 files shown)


sildater:
Thank you for the quick fix! I added some minor details to the estimation but I think it works nicely now. Since I added code, feel free to review and merge.

manoskary (commented):
The Straight Durs is a nice global that fixes a lot of potential issues, I agree that it is a crucial addition for making the estimation more robust. The last changes by @sildater also include some better styling and more detailed warnings about composite symbolic durations.

partitura/utils/globals.py
```diff
@@ -192,6 +192,24 @@
     {"type": "long", "dots": 3},
 ]
 
+STRAIGHT_DURS = np.array(
```
fosfrancesco, line 195: I would add a comment with a high level explanation about why this is useful! Now it looks a bit obscure.

fosfrancesco (changes requested):

fosfrancesco (approved):

PR #382, opened 2024-10-02
Clef feature

sildater:
added a simple note feature encoding the currently active clef by sign, line, and octave change.


tests/test_clef.py
```diff
@@ -0,0 +1,22 @@
...
+
+class TestingClefFeatureExtraction(unittest.TestCase):
+    def test_feature_exctraction(self):
+        for fn in CLEF_TESTFILES:
+            score = load_musicxml(fn)
+            sna1 = compute_note_array(score.parts[0],
+                               feature_functions=["clef_feature"])
+            sna2 = compute_note_array(score.parts[1],
+                               feature_functions=["clef_feature"])
+
```
manoskary, line 22: I have a couple of questions about this segment:
- Is this test finished? 
- why are you calling the compute_note_array on individual parts instead of the score?
- Would you like to add an assertion at the end validating the values?

partitura/musicanalysis/note_features.py
```diff
@@ -526,6 +527,70 @@ def grace_feature(na, part, **kwargs):
...
+            interpolator = interp1d(start_times, clef_indices, 
+                                    kind = "previous", 
+                                    bounds_error=False, 
+                                    fill_value="extrapolate")
+            clef_dict[interpolator_key].append(interpolator)
+
+        W = np.zeros((len(notes), 3))
+
+        for i, n in enumerate(notes):
+            staff = n.staff or 1
```
manoskary, line 573: By iterating `notes = part.notes_tied` there is no guarantee that the order of the note/features will match the one in `note_array`. Maybe you can add a double check about this where you check if the `n.id == na[i]["id"]` and if not the same then seek the correct note to update the features so it follows the order of the `note_array`.

manoskary (changes requested):
Very nice addition, this would be a very useful feature, there are a couple of comments that need to be addressed before merging this PR.

tests/test_clef.py
sildater: is not finished

manoskary: Ok, I will wait for you @sildater  to re-request review when ready!

sildater: @manoskary is ready now

manoskary (changes requested):
This PR adresses a Clef feature within note features, it also corrects a number of vulnerabilities at other note_features dependent on the notes_tied instead of the note_array, To ensure correct order a remapping was proposed using a dictionary with note ids, however there is a small comment regarding the uniqueness of note ids.

partitura/musicanalysis/note_features.py
```diff
@@ -526,6 +527,71 @@ def grace_feature(na, part, **kwargs):
...
 
 
+def clef_feature(na, part, **kwargs):
+    """Clef feature
+
+    This feature encodes the current clef of the staff of each note.
+    Note that this feature does not return the staff number per note,
+    see staff_feature for this information.
+    """
+    notes = {n.id:n for n in part.notes_tied}
```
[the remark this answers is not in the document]
sildater: if note or rest IDs are not unique, the value of the last note of that ID will be used in in the output array. That's wrong, but due to the data and non-breaking.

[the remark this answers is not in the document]
manoskary: In that case, maybe adding a warning would be nice about the feature.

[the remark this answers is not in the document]
sildater: added two warnings in the top level feature functions

manoskary (approved):

PR #387, opened 2024-10-10
Relax assertion tuplet

manoskary:
- Relaxing assertion for Tuplets to `start_tuplet.start_note.start.t <= stop_tuplet.end_note.start.t` to avoid error with slightly problematic scores
- Relax conditions for ChordSymbols by accepting empty kind. Sometimes, some instructions in the asap-dataset are pushed like harmony labels so, in the future, a test should ideally filter them out in partitura.


sildater (approved):
I didn't test it but it seems reasonable to avoid errors for spurious single note tuplets.

PR #391, opened 2024-10-17
patch issue 390

sildater:
BUG: cost-based merging of non-note elements into stream leads to attributes not being in the beginning of the measure, which in turn leads to problems with musescore.
FIX: add attributes first, then the notes. I also added a comment to document the change in place as I'm unsure about downstream consequences but no tests fail.


manoskary (approved):
Approved changes as they do not seem to affect the performance, asked second opinion from @CarlosCancino-Chacon

PR #392, opened 2024-10-18
Stem Direction Support

manoskary:
- Added support for stem direction in musicxml import/export
- Added support for stem direction export in mei export
- Added a generic note attribute called beam_direction initialized to None including minimal tests to only accept "up/down" values.

### DIFF
--- a/tests/data/musicxml/test_unfold_complex_result.xml
+++ b/tests/data/musicxml/test_unfold_complex_result.xml
@@ -57,6 +57,7 @@
         </pitch>
         <duration>1</duration>
         <voice>1</voice>
+        <stem>up</stem>
         <type>quarter</type>
       </note>
       <note>
@@ -66,6 +67,7 @@
         </pitch>
         <duration>1</duration>
         <voice>1</voice>
+        <stem>up</stem>
         <type>quarter</type>
       </note>
     </measure>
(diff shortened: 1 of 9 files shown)


tests/data/musicxml/test_unfold_complex_result.xml
sildater, line 1: this file is not used in the tests

tests/data/musicxml/test_unfold_dacapo_result.xml
sildater, line 1: this files is not used in the test

tests/data/musicxml/test_unfold_volta_numbers_result.xml
sildater, line 1: this file is not used in the test

tests/test_xml.py
sildater, line 1: the test covers a single note of one file in one function.

sildater (approved):
Nice and simple stem encoding addition! There is probably a mistake in the original message, there is no change related to beams that I could see, only stems. Nearly all testing data changes are unused but also don't break anything, and the test is as minimal as it gets, but it's there.

tests/data/musicxml/test_unfold_complex_result.xml
manoskary: Hello @sildater. This file caused an error because since stem direction is now supported (and some were present in the original unfolded scores), it created a diff error with the unfolded variant on a different test. The same happened for all the other files you commented on below.

PR #394, opened 2024-10-19
Fix typo: importnakamura.py

mimbres:
fix typo, match.txt --> corresp.txt

### DIFF
--- a/partitura/io/importnakamura.py
+++ b/partitura/io/importnakamura.py
@@ -35,7 +35,7 @@ def load_nakamuracorresp(filename: PathLike) -> Tuple[Union[np.ndarray, list]]:
     Parameters
     ----------
     filename : str
-        The nakamura match.txt-file
+        The nakamura corresp.txt-file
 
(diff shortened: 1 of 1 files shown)


manoskary:
this is not a typo, the file is named match.

mimbres:
Could you check again? 
- I noticed that the [`nakamura` alignment algorithm](https://midialignment.github.io/demo.html) v240109 (C++) outputs both `match.txt` and `corresp.txt`. 
- In `Partitura`, `pt.load_nakamuramatch` uses the `match.txt` file as input. (current document is correct)
- However, for `pt.load_nakamuracorresp`, using the `match.txt` file causes an error. You must use the `corresp.txt` file instead—**the documentation is incorrect.**

manoskary:
Hello @mimbres, I'm sorry for the early dismissal. You are right; there is probably a typo on the document. However, since this is a very small documentation edit, I will go ahead and correct it in one of the already open PRs.

PR #399, opened 2024-11-05
Performance pedal fix

huispaty:
This PR includes a pedal fix for notes with reonset(s) during sustain pedal when creating performance note arrays from MIDI files.

### DIFF
--- a/partitura/performance.py
+++ b/partitura/performance.py
@@ -178,7 +178,8 @@ class PerformedPart(object):
             duration_sec = offset - note_on_sec
             duration_tick = (
                 n.get(
-                    "note_off_tick",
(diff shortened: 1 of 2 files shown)


CarlosCancino-Chacon (approved):

CarlosCancino-Chacon:
This pull request provides a bug fix when adjusting the note onsets/offsets (and sound off) due to pedal changes.

* It fixes the duration in ticks of the sound off (previously, only the duration in seconds was adjusted with pedal information, and the corresponding duration in ticks was not updated)
* A fix for the correct adjusted note offset considering new notes (i.e., avoid overlapping notes of the same pitch)

PR #404, opened 2024-12-03
Load Scores from URL

manoskary:
This pull request introduces several enhancements to the `partitura` library, including the addition of URL handling for score loading and updates to interval definitions. Additionally, new unit tests have been added to validate these changes.

Enhancements to score loading:

* [`partitura/io/__init__.py`](diffhunk://#diff-4950763a208c8a9eed6ce66988ff2bfc439dabbb3fd249327f4fd05008b7e925L8-R10): Added functions `is_url` and `download_file` to handle URL inputs, and modified the `load_score` function to support loading scores from URLs. 

Updates to interval definitions:

* [`partitura/score.py`](diffhunk://#diff-f38b43739939f124faf420e4ede71c9ab2e32542919b48d6ee33da924cd13e73L6040-R6041): Corrected the intervals for "vi" and "vii" from minor to major in the `process_local_key` function.

New unit tests:

* [`tests/test_urlload.py`](diffhunk://#diff-1f9efff094c53ac978bcfd587a636fc08aaf5ca0e6b94303fc2e4f51d7913177R1-R26): Added new unit tests to ensure that scores can be loaded from URLs for various file formats, including Kern, MEI, MIDI, and MusicXML.


partitura/io/__init__.py
```diff
@@ -32,6 +34,32 @@ class NotSupportedFormatError(Exception):
...
+def download_file(url):
+    # Send a GET request to the URL
+    with urllib.request.urlopen(url) as response:
+        data = response.read()
+
+    # Extract the file extension from the URL
+    extension = os.path.splitext(url)[-1]
+
+    # Create a temporary file
+    temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=extension)
```
fosfrancesco, line 54: Won't it be the case that this file is never deleted, and keep accumulating on the computer?

fosfrancesco (changes requested):

partitura/io/__init__.py
manoskary: Yes, I agree.  Probably, I need to find a better solution and debug. The load score file needs access to the file to check the extension but maybe when `delete=True` the file is not deleted immediately when loaded. I will push a change soon.

manoskary: Turns out that for python version 3.8 the temporary file options are not the same as for newer versions of python therefore I had to add more edits to ensure the deletion of tmp files immediately after the end of the function.

fosfrancesco (approved):

Issue #406, opened 2024-12-18
Issue with MuseScore import with chord at tuplet start


leleogere:
I'm playing around with score generation, including tuplets and chords. However, I'm facing an issue when importing the score into MuseScore after that, as it thinks it's corrupted even if the score should be valid. It seems to have trouble parsing a tuplet when the first note is a chord, and the `<tuplet>` tag is not on the first note of the chord in the MusicXML file.

Here is the badly imported score:

![Original MusicXML](https://github.com/user-attachments/assets/f00992e0-ccb6-49a9-87ac-c4af38694ae7)

When it should look like this:

![Fixed MusicXML](https://github.com/user-attachments/assets/04bf1a4f-9b90-4a9a-b733-b3765307dc9e)

To get this second image, I simply moved the `<tuplet>` part from the second note of the chord to the first one:
<details>
<summary>Show diff</summary>

```diff
 <?xml version='1.0' encoding='UTF-8'?>
 <!DOCTYPE score-partwise PUBLIC
   "-//Recordare//DTD MusicXML 3.1 Partwise//EN"
   "http://www.musicxml.org/dtds/partwise.dtd">
 <score-partwise>
   <part-list>
     <score-part id="P1">
       <part-name/>
     </score-part>
   </part-list>
   <part id="P1">
     <!--=======================================================-->
     <measure>
       <attributes>
         <divisions>480</divisions>
         <time>
           <beats>1</beats>
           <beat-type>4</beat-type>
         </time>
       </attributes>
       <note>
         <pitch>
           <step>A</step>
           <alter>-2</alter>
           <octave>5</octave>
         </pitch>
         <duration>252</duration>
         <type>eighth</type>
         <dot/>
         <dot/>
         <time-modification>
           <actual-notes>5</actual-notes>
           <normal-notes>3</normal-notes>
         </time-modification>
+        <notations>
+          <tuplet number="1" type="start"/>
+        </notations>
       </note>
       <note>
         <chord/>
         <pitch>
           <step>D</step>
           <alter>-1</alter>
           <octave>5</octave>
         </pitch>
         <duration>252</duration>
         <type>eighth</type>
         <dot/>
         <dot/>
         <time-modification>
           <actual-notes>5</actual-notes>
           <normal-notes>3</normal-notes>
         </time-modification>
-        <notations>
-          <tuplet number="1" type="start"/>
-        </notations>
       </note>
       <note>
         <chord/>
         <pitch>
           <step>B</step>
[...]

PR #407, opened 2024-12-19
Sort exported notes by step when same `midi_pitch`

leleogere:
In MusicXML export, the notes are currently sorted using their `midi_pitch` attribute. However, this can lead to inconsistent exports when the score contains notes that are [enharmonic equivalents](https://en.wikipedia.org/wiki/Enharmonic_equivalence) (such as D# and Eb, or B# and C), as the export will differ depending on the order the user inserts the notes in the `Part` object:
```py
score.add(pt.score.Note("C", 4, 0), start=0, end=1)  # midi_pitch=60
score.add(pt.score.Note("B", 3, 1), start=0, end=1)  # midi_pitch=60
# will result in a different file than
score.add(pt.score.Note("B", 3, 1), start=0, end=1)  # midi_pitch=60
score.add(pt.score.Note("C", 4, 0), start=0, end=1)  # midi_pitch=60
```

This PR simply adds the step as a second key for the sorting. Therefore, notes are first sorted by `midi_pitch`, and when they are enharmonic equivalents they are sorted by `step`, so that the export file is consistent no matter the order the notes are added to the score.

This will solve a potential issue raised in the second message of #406.

### DIFF
--- a/partitura/io/exportmusicxml.py
+++ b/partitura/io/exportmusicxml.py
@@ -417,9 +417,10 @@ def linearize_segment_contents(part, start, end, state):
 
     for voice in sorted(notes_by_voice.keys()):
         voice_notes = notes_by_voice[voice]
-        # sort by pitch
+        # sort by pitch (then step in case of enharmonic notes)
         voice_notes.sort(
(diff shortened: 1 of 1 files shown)


fosfrancesco:
An effective solution for the inconsistent ordering of enharmonic notes in the musicxml export.

PR #410, opened 2025-01-10
option to to force new segments in `add_segments`

sildater:
`add_segments` used to do nothing when segments are present, but several unfolding methods read and overwrite segment attributes in place, which renders them unusable for further unfolding/processing. now one can start from scratch using the flag `force_new` in `add_segments` to get freshly initialized segments. the default behavior doesn't change.


huispaty (approved):
The new argument `force_new` allows creating new unfoldings if segment markers have been changed, which makes score processing further down the line more convenient and flexible.

Issue #405, opened 2024-12-16
Direct access to `actual_notes`, `normal_notes` and `type` in a tuplet


leleogere:
Currently, tuplets objects only consist of pointers to the start and end note of the tuplet. There is no way of saying if it is a "3-against-2-eights" triplet for example, or a "5-against-4-sixteenths" quintuplet when iterating over the score and encountering a `Tuplet` object. To get this information, one would have to explore the notes in between the start and end note, trying to guess this information from their type (which is not straightforward as a type `quarter_3/2` could be either a quarter note in an eighth-note triplet or a quarter note in a quarter-note tuplet for example, or for example when parsing nested tuplets).

The actual/normal notes information is sometimes accessible in the MusicXML file. For example, below is an example of a note on which are starting 2 nested tuplets: a 3-against-2-eight triplet and a 3-against-2-sixteenth triplet.
<details>
<summary>Show XML</summary>

```xml
      <note>
        <pitch>
          <step>E</step>
          <octave>4</octave>
          </pitch>
        <duration>1</duration>
        <voice>1</voice>
        <type>16th</type>
        <time-modification>
          <actual-notes>9</actual-notes>
          <normal-notes>4</normal-notes>
          </time-modification>
        <stem>up</stem>
        <beam number="1">begin</beam>
        <beam number="2">begin</beam>
        <notations>
          <tuplet type="start" number="1" bracket="yes" placement="below">
            <tuplet-actual>
              <tuplet-number>3</tuplet-number>
              <tuplet-type>eighth</tuplet-type>
              </tuplet-actual>
            <tuplet-normal>
              <tuplet-number>2</tuplet-number>
              <tuplet-type>eighth</tuplet-type>
              </tuplet-normal>
            </tuplet>
          <tuplet type="start" number="2" bracket="yes">
            <tuplet-actual>
              <tuplet-number>3</tuplet-number>
              <tuplet-type>16th</tuplet-type>
              </tuplet-actual>
            <tuplet-normal>
              <tuplet-number>2</tuplet-number>
              <tuplet-type>16th</tuplet-type>
              </tuplet-normal>
            </tuplet>
          </notations>
        </note>
``` 
</details>

However, this information is not always explicitly provided (in scores exported by MuseScore at least), and is supposed to be inferred from the `<time-modification>` tags if absent (according to [MusicXML spec](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/tuplet-actual)).
<details>
[...]

PR #414, opened 2025-01-17
Add optional attributes to tuplets related to their time modification

leleogere:
As I explained in #405, I thought that tuplets objects were missing some information about the rhythmic modifications they induce (actual notes, normal notes and type). In MusicXML scores, tuplets elements *can* contain this information in [`<tuplet-actual>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/tuplet-actual) and [`<tuplet-normal>`](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/tuplet-normal) tags, but usually, they do not need to be specified and can be inferred from the `<time-modification>` tags of notes. Actually, MuseScore only exports such tags in case of nested tuplets as `<time-modification>` tags result from the multiplication of the different nested ratio.

Currently, Partitura does not parse those tags, and therefore some information is lost in case of nested tuplets. Even in the case of simple single-level tuplets, I feel that this information should be accessible at the `Tuplet` object level, in addition to the note level.

This PR allows exactly that, by adding optional `actual_notes`, `normal_notes` and `type` attributes, that are parsed from the MusicXML `<tuplet-actual/normal>` tags if present, and otherwise inferred from the first note of the tuplet.

This also adds a property `duration_multiplier` to the `Tuplet` similarly to what music21 proposes (which is simply `normal_notes/actual_notes`).

I also implemented it at export time, always outputting the tags `<tuplet-actual/normal>` no matter if they are absolutely needed or only optional, I don't think that it could cause any issue?

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -1492,36 +1492,41 @@ def handle_tuplets(notations, ongoing, note):
                 tuplet_actual_type = get_value_from_tag(tuplet_actual, "tuplet-type", str)
                 tuplet_normal_notes = get_value_from_tag(tuplet_normal, "tuplet-number", int)
                 tuplet_normal_type = get_value_from_tag(tuplet_normal, "tuplet-type", str)
-                # Types should always be the same I think?
-                assert tuplet_actual_type == tuplet_normal_type, "Tuplet types are not the same"
-                tuplet_type = tuplet_actual_type
             # If no information, try to infer it from the note
             else:
                 tuplet_actual_notes = note.symbolic_duration.get("actual_notes", None)
                 tuplet_normal_notes = note.symbolic_duration.get("normal_notes", None)
-                tuplet_type = note.symbolic_duration.get("type", None)
+                tuplet_actual_type = note.symbolic_duration.get("type", None)
+                tuplet_normal_type = tuplet_actual_type
 
             # If anyone of the attributes is not set, we set them all to None as we can't really
             # do anything useful with only partial information about the tuplet
-            if None in (tuplet_actual_notes, tuplet_normal_notes, tuplet_type):
+            if None in (tuplet_actual_notes, tuplet_normal_notes, tuplet_actual_type, tuplet_normal_type):
                 tuplet_actual_notes = None
                 tuplet_normal_notes = None
-                tuplet_type = None
-
+                tuplet_actual_type = None
+                tuplet_normal_type = None
 
             # check if we have a stopped_tuplet in ongoing that corresponds to
             # this start
             tuplet = ongoing.pop(stop_tuplet_key, None)
 
             if tuplet is None:
-                tuplet = score.Tuplet(note, actual_notes=tuplet_actual_notes, normal_notes=tuplet_normal_notes, type=tuplet_type)
+                tuplet = score.Tuplet(
+                    note,
+                    actual_notes=tuplet_actual_notes,
+                    normal_notes=tuplet_normal_notes,
+                    actual_type=tuplet_actual_type,
+                    normal_type=tuplet_normal_type,
+                )
                 ongoing[start_tuplet_key] = tuplet
 
             else:
                 tuplet.start_note = note
                 tuplet.actual_notes = tuplet_actual_notes
                 tuplet.normal_notes = tuplet_normal_notes
-                tuplet.type = tuplet_type
+                tuplet.actual_type = tuplet_actual_type
+                tuplet.normal_type = tuplet_normal_type
 
             starting_tuplets.append(tuplet)
 
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -698,7 +698,10 @@ def _handle_harmony(e, position, part):
         text = e.find("function").text
         if text is not None:
             if "|" in text:
-                text, cadence_annotation = text.split("|")
+                text = text.split("|")
+                if len(text) > 2:
+                    warnings.warn(f"Ignoring multiple cadence annotations {text[2:]}", stacklevel=2)
+                text, cadence_annotation = text[0], text[1]
                 part.add(score.Cadence(cadence_annotation), position)
             part.add(score.RomanNumeral(text), position)
     elif e.find("kind") is not None and e.find("root") is not None:
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -1441,16 +1441,48 @@ def handle_tuplets(notations, ongoing, note):
         stop_tuplet_key = ("stop_tuplet", tuplet_number)
 
         if tuplet_type == "start":
+            # Get information about the tuplet if present in the XML
+            tuplet_actual = tuplet_e.find("tuplet-actual")
+            tuplet_normal = tuplet_e.find("tuplet-normal")
+            if tuplet_actual is not None and tuplet_normal is not None:
+                tuplet_actual_notes = int(tuplet_actual.find("tuplet-number").text)
+                tuplet_actual_type = tuplet_actual.find("tuplet-type").text
+                tuplet_normal_notes = int(tuplet_normal.find("tuplet-number").text)
+                tuplet_normal_type = tuplet_normal.find("tuplet-type").text
+                # Types should always be the same I think?
+                assert tuplet_actual_type == tuplet_normal_type, "Tuplet types are not the same"
+                tuplet_type = tuplet_actual_type
+
+            # If no information, try to infer it from the note
+            elif (
+                "actual_notes" in note.symbolic_duration
+                and "normal_notes" in note.symbolic_duration
+                and "type" in note.symbolic_duration
+                ):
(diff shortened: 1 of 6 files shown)


partitura/score.py
```diff
@@ -2444,10 +2448,17 @@ def end_note(self, note):
             note.tuplet_stops.append(self)
         self._end_note = note
 
+    @property
+    def duration_multipler(self) -> Fraction:
+        return Fraction(self.normal_notes, self.actual_notes)
```
manoskary, line 2453: I think there is a typo in this property. Also, is it used anywhere?

partitura/io/importmusicxml.py
```diff
@@ -1441,16 +1441,48 @@ def handle_tuplets(notations, ongoing, note):
...
+                    tuplet_actual_notes = note.symbolic_duration["actual_notes"]
+                    tuplet_normal_notes = note.symbolic_duration["normal_notes"]
+                    tuplet_type = note.symbolic_duration["type"]
+
+            # If no information is present in the XML or the note, then set to None
+            else:
+                tuplet_actual_notes = None
+                tuplet_normal_notes = None
+                tuplet_type = None
+
```
manoskary, line 1471: Could you do it with one clause to cover all combinations, or would it be wrong? i.e.:
```python
tuplet_actual_notes = note.symbolic_duration["actual_notes"] if "actual_notes" in note.symbolic_duration else None
tuplet_normal_notes = note.symbolic_duration["normal_notes"] if "normal_notes" in note.symbolic_duration else None
tuplet_type = note.symbolic_duration["type"] if "type" in note.symbolic_duration else None
```

partitura/io/importmusicxml.py
```diff
@@ -1441,16 +1441,48 @@ def handle_tuplets(notations, ongoing, note):
...
 
         if tuplet_type == "start":
+            # Get information about the tuplet if present in the XML
+            tuplet_actual = tuplet_e.find("tuplet-actual")
+            tuplet_normal = tuplet_e.find("tuplet-normal")
+            if tuplet_actual is not None and tuplet_normal is not None:
+                tuplet_actual_notes = int(tuplet_actual.find("tuplet-number").text)
+                tuplet_actual_type = tuplet_actual.find("tuplet-type").text
+                tuplet_normal_notes = int(tuplet_normal.find("tuplet-number").text)
+                tuplet_normal_type = tuplet_normal.find("tuplet-type").text
```
manoskary, line 1451: Instead of calling `attr.find`, I recommend using the function `get_value_from_attribute`, which is slightly more flexible to cover missing values.

manoskary (changes requested):
Very nice addition to Partitura to cover better support of tuplets in musicxml . Review contains only minor comments, suggestions, and questions

partitura/score.py
leleogere: Yes, I'll fix the typo. It is not used anywhere so far, it's a utility inspired by the [`tupletMultiplier`](https://www.music21.org/music21docs/moduleReference/moduleDuration.html#music21.duration.Tuplet.tupletMultiplier) property of music21 and I thought that it might be useful, however I can remove it if you think this does not belong to partitura

manoskary: No, it could be nice to have it. Maybe add some documentation on the property and a link to the music21 property so if somebody wants to use it in the future it would be easier to know how.

partitura/io/importmusicxml.py
leleogere: I thought about that, but I'm not sure what would be the right approach. When you only have access to part of the information, you are quite limited in what you can do with the partial information. Especially, at export time, you can't export this partial information (I haven't tried to do it to see how MuseScore would behave, but I guess that I would trigger the corrupted file popup).

Therefore, I chose to simply not fill the attribute when one of them was missing (I think it should not happen, but you never know with those funky musicxml around). If you think that it make sense to fill them even when partially present, I can add that. But in that case, how should I cover the export? Should I add the `<tuplet-normal/actual>` tags only when the 3 attributes are filled, and drop them otherwise? Or try to fill them partially anyway and let the problem for whatever musicxml reader people use?

partitura/score.py
leleogere: I'll do that!

partitura/io/importmusicxml.py
leleogere: Maybe I'm wrong, but doesn't `get_value_from_attribute` only work for attributes of a tag? The `<tuplet>` tag is organised as follow:

```xml
        <notations>
          <tuplet type="start" number="2" bracket="no">
            <tuplet-actual>
              <tuplet-number>3</tuplet-number>
              <tuplet-type>16th</tuplet-type>
              </tuplet-actual>
            <tuplet-normal>
              <tuplet-number>2</tuplet-number>
              <tuplet-type>16th</tuplet-type>
              </tuplet-normal>
            </tuplet>
          </notations>
```

And `<tuplet-actual/normal>` are not attributes of `<tuplet>` but rather subelements, hence the `.find` method

partitura/io/importmusicxml.py
manoskary: Good question. I am not entirely sure either. I think you should choose the one that feels more correct to you on this, but just in case add some inline comments explaining this choice if in the future there is a need for changing.

partitura/io/importmusicxml.py
manoskary: Yes you are right, I meant to write: `get_value_from_tag`

leleogere: Oh yes I missed this one!

partitura/io/importmusicxml.py
leleogere: I think I'll leave it as is and add some comments

leleogere:
I case you have planned to merge this right now, do not yet, I've found an issue I would like some insights about. I'll explain in a comment on the related code in a few minutes.

leleogere:
The issue I mentioned is about the following assert:
https://github.com/leleogere/partitura/blob/ab426ee43cb957b2a23ab17f8c5deb60a55b5fed/partitura/io/importmusicxml.py#L1453
```python
assert tuplet_actual_type == tuplet_normal_type, "Tuplet types are not the same"
```

 Initially, I thought that the `<tuplet-type>` would always be the same in `<tuplet-actual>` and `<tuplet-normal>`, as tuplets are usually 3 eighths in the space of 2 eighths, or 5 quarters in the space of 4 quarters... I couldn't really see a case where we would like to do like 3 eights in the space of 2 quarters, therefore, I assumed that this assert should be ok. 

However, the tag belonging to `<tuplet-actual>` AND `<tuplet-normal>` should have been an insight that the MusicXML spec allows them to be different, and I actually found an example of such a piece: 
https://github.com/CPJKU/asap-dataset/blob/main/Chopin/Ballades/1/xml_score.musicxml (lines 79875 to 79884)
```xml
          <tuplet number="1" placement="below" type="start">
            <tuplet-actual>
              <tuplet-number>13</tuplet-number>
              <tuplet-type>eighth</tuplet-type>
            </tuplet-actual>
            <tuplet-normal>
              <tuplet-number>4</tuplet-number>
              <tuplet-type>quarter</tuplet-type>
            </tuplet-normal>
          </tuplet>
```
This piece contains a lot of occurrences of `tuplet_actual_type != tuplet_normal_type` (simply search for `tuplet-type` in the xml), and therefore will trigger the `AssertionError` when parsed by Partitura.

The way MuseScore deals with it seems to be to convert the `<tuplet-normal>` unit into the one in `<tuplet-actual>` one. Therefore, in that case, it interprets the `normal` part as 8 eighth notes rather than 4 quarter notes as specified in the XML:
image

Therefore here is my question, should I separate the `type` attribute of tuplets into two attributes `normal_type` and `actual_type` as specified in the xml? Or rather convert the `normal-type` in the same type as the `actual_type` as MuseScore does and keep a single `type` attribute as I've done now?

I would be interested in your insights about this issue @manoskary

manoskary:
Hello Leo and thanks for the detailed message. It is true that we tend to find the weirdest things in score encodings and this would not be only example for sure. Generally, I would recommend using the option that would not break the parser. In this case I think separating the type could actually be a good solution for the import, in order to respect the encoding. For the export, I would use the MuseScore solution so it does not break. As always add some inline comments explaining why. Would this solution sound reasonable to you?

leleogere:
I've split the `.type` attribute into two `.actual_type` and `.normal_type` attribute. In the export, I've exported them as is, but let me know if you would rather do the conversion `normal-type` to `actual-type` anyway at export time. I had to deal with the conversion anyway in `.duration_multiplier` as it needs to be taken into account in case of `actual_type != normal_type`.

manoskary:
Ok thanks let me run some tests before merging

manoskary (approved):
After running some tests on bigger Musicxml sets, loading and exporting is not failing therfore I will move forward with merging this PR. If in the future we notice, incorrect visualizations of tuplets, maybe we can rethink the export part.

Issue #415, opened 2025-01-17
Error Loading Specific MusicXML Files in Composer Identification Baseline


AidaRostamza:
Hello,

We are writing to report an issue with the Baseline_ComposerIdentification.py script when processing specific MusicXML files. The following files cause an error:

- 275.musicxml
- 002.musicxml
- 003.musicxml

The error occurs at this line:

score = pt.load_musicxml(fn)

We have attached a screenshot of the error for your reference.

Best Regards
Team Gyorgy Ligeti

![Image](https://github.com/user-attachments/assets/b5082946-9ddd-4b22-a66b-fecba928d37c)

manoskary:
Thank you for reporting this issue. We will propose a fix as soon as possible!

PR #416, opened 2025-01-18
Fix Issue #414

CarlosCancino-Chacon:
This pull request fixes issue #415.

Some MusicXML files have incorrectly formatted fingering information (the value of the attribute `<fingering>`  should be an integer, but some files include other strings such as `4_1`).

In the current solution, parsing these strings would only take the first value and would ignore subsequent values. Otherwise, if the string cannot be parsed, the fingering would be set to None, instead of raising an error.


manoskary (approved):
Updates on fingering extended options and better parsing of annotations using regex

PR #417, opened 2025-01-20
Replaced deprecated scipy derivative

manoskary:
This pull request introduces a new function `first_order_derivative` and replaces the usage of `scipy.misc.derivative` with this new function in the `partitura` package. As of scipy version 1.15 the function  `scipy.misc.derivative` is deprecated.

### DIFF
--- a/partitura/utils/generic.py
+++ b/partitura/utils/generic.py
@@ -174,6 +174,8 @@ def iter_subclasses(cls, _seen=None):
 def first_order_derivative(func, x0, dx=0.5):
     """
     Compute the first order derivative of a function at a point `x0` using
+    a central difference scheme. This function is an adaptation of the scipy _derivative
+    function, which is no longer available in the public API after version 1.15.
     Parameters
     ----------
     func: callable
@@ -188,8 +190,6 @@ def first_order_derivative(func, x0, dx=0.5):
     np.ndarray
         First order derivative of `func` at `x0`
     """
-    # fixed order
-    n = 1
     # fixed number of points to use, must be odd.
     num_points = 3
     weights = np.array([-1, 0, 1]) / 2.0
@@ -197,7 +197,7 @@ def first_order_derivative(func, x0, dx=0.5):
     ho = num_points >> 1
     for k in range(num_points):
         val += weights[k] * func(x0 + (k - ho) * dx)
-    return val / np.prod((dx,) * n, axis=0)
+    return val / np.prod((dx,), axis=0)
 
 
 class ReplaceRefMixin(object):
--- a/partitura/utils/generic.py
+++ b/partitura/utils/generic.py
@@ -171,6 +171,35 @@ def iter_subclasses(cls, _seen=None):
                 yield sub
 
 
+def first_order_derivative(func, x0, dx=0.5):
+    """
+    Compute the first order derivative of a function at a point `x0` using
+    Parameters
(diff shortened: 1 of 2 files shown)


partitura/utils/generic.py
```diff
@@ -171,6 +171,35 @@ def iter_subclasses(cls, _seen=None):
                 yield sub
 
 
+def first_order_derivative(func, x0, dx=0.5):
+    """
+    Compute the first order derivative of a function at a point `x0` using
```
fosfrancesco, line 176: mention here where this code was taken from, so people can check it up

partitura/utils/generic.py
```diff
@@ -171,6 +171,35 @@ def iter_subclasses(cls, _seen=None):
...
+    # fixed order
+    n = 1
+    # fixed number of points to use, must be odd.
+    num_points = 3
+    weights = np.array([-1, 0, 1]) / 2.0
+    val = 0.0
+    ho = num_points >> 1
+    for k in range(num_points):
+        val += weights[k] * func(x0 + (k - ho) * dx)
+    return val / np.prod((dx,) * n, axis=0)
```
fosfrancesco, line 200: this can probably be simplified, since n is always 1

fosfrancesco (changes requested):

partitura/utils/generic.py
manoskary: You're right I will do that

partitura/utils/generic.py
```diff
@@ -171,6 +171,35 @@ def iter_subclasses(cls, _seen=None):
...
+        First order derivative of `func` at `x0`
+    """
+    # fixed number of points to use, must be odd.
+    num_points = 3
+    weights = np.array([-1, 0, 1]) / 2.0
+    val = 0.0
+    ho = num_points >> 1
+    for k in range(num_points):
+        val += weights[k] * func(x0 + (k - ho) * dx)
+    return val / np.prod((dx,), axis=0)
```
fosfrancesco, line 200: np.prod((dx,), axis=0) can be also simplified in dx probably?

partitura/utils/generic.py
```diff
@@ -171,6 +171,35 @@ def iter_subclasses(cls, _seen=None):
                 yield sub
 
 
+def first_order_derivative(func, x0, dx=0.5):
```
fosfrancesco, line 174: the default dx in the scipy function is 1. Is this different value on purpose?

partitura/utils/generic.py
```diff
@@ -171,6 +171,35 @@ def iter_subclasses(cls, _seen=None):
...
+    """
+    Compute the first order derivative of a function at a point `x0` using
+    a central difference scheme. This function is an adaptation of the scipy _derivative
```
fosfrancesco, line 177: I would add the link https://github.com/scipy/scipy/blob/v1.7.0/scipy/misc/common.py#L75-L145 
which should stay the same

fosfrancesco (commented):
Sorry, more comments. But they should be fast

partitura/utils/generic.py
manoskary: This is just the default value that was used in tempo by derivate which was the use of this function in first place. I am just propagating the default also here.

fosfrancesco (approved):

PR #418, opened 2025-01-21
equal duration for all chord notes

huispaty:
Use the duration of the first chord note as the duration for all remaining notes of that chord if their duration is unequal.


sildater (approved):

sildater:
This PR addresses an issue that came up during the parsing of the [BPSD data](https://zenodo.org/records/12783403): not all notes in a musicxml \chord group shared the same duration, which shifts the timeline and leads to all manner of strange downstream problems because \chord objects belong to the same voice. This fix assumes this to be an encoding error and sets the duration of all elements to the duration of the first.

Issue #420, opened 2025-01-27
Tuplet parsing issue in measure 125 of `Chopin/Sonata_2/1st_no_repeat`


leleogere:
Here are the measures 125 and 126 of [Chopin/Sonata_2/1st_no_repeat](https://github.com/CPJKU/asap-dataset/blob/main/Chopin/Sonata_2/1st_no_repeat/xml_score.musicxml):

![Image](https://github.com/user-attachments/assets/bc030cd3-d683-4f30-9686-cd5d38d8ce90)

They each contain 3 tuplets, two on voice 5 and one on voice 6. However, when parsing the score with the following code, we can see that only two of them:
```py
from pathlib import Path

import partitura as pt
import partitura.score

score_path = Path("asap-dataset/Chopin/Sonata_2/1st_no_repeat/xml_score.musicxml")

score = pt.load_musicxml(score_path)
score = score.parts[0]

measure_map = score.measure_number_map
printed_measures = set()  # To avoid printing the measure number each time

for element in score.iter_all(cls=pt.score.Tuplet):
    # Print measure number for first element of measure
    if (measure := int(measure_map(element.start.t))) not in printed_measures:
        print(f"Measure {measure}")
        printed_measures.add(measure)
    print(f"  * {str(element)}")
    print(f"      - start_note = {str(element.start_note)}")
    print(f"      - end_note   = {str(element.end_note)}")

```
```
[...]
Measure 125
  * 11916--11964 Tuplet start=None end=None actual_notes=3 normal_notes=2 actual_type=quarter normal_type=quarter
      - start_note = 11916--11932 Note id=None voice=5 staff=2 type=quarter_3/2 pitch=E2
      - end_note   = 11948--11964 Note id=None voice=5 staff=2 type=quarter_3/2 pitch=G3
  * 11964--12012 Tuplet start=None end=None actual_notes=3 normal_notes=2 actual_type=quarter normal_type=quarter
      - start_note = 11964--11980 Note id=None voice=5 staff=2 type=quarter_3/2 pitch=F#3
      - end_note   = 11996--12012 Note id=None voice=5 staff=2 type=quarter_3/2 pitch=D2
Measure 126
  * 12012--11964 Tuplet start=None end=None actual_notes=3 normal_notes=2 actual_type=quarter normal_type=quarter
      - start_note = 12012--12028 Note id=None voice=5 staff=2 type=quarter_3/2 pitch=F2
[...]

PR #423, opened 2025-01-28
Add a warning when tuplet.start_note.voice != tuplet.end_note.voice

leleogere:
As discussed in #420, the start and end notes of a tuplet are a good indicator of a malformed XML error. This happens when there is a missing tag `<tuplet type="start">` or `<tuplet type="stop">`. For instance, this happens in the score Chopin/Sonata_2/1st_no_repeat in ASAP (see fix at https://github.com/fosfrancesco/asap-dataset/pull/14).

This PR adds a warning when this happens, providing information to the user that there might be a missing tuplet tag.
```py
/home/user/partitura/partitura/io/importmusicxml.py:1577: UserWarning: Tuplet start and end notes do not belong to the same voice (5 != 6). This might indicate a missing <tuplet type="start"> or <tuplet type="stop">.
  warnings.warn(
```

MuseScore is able to fix this kind of encoding error, should we try too? Here is an overview of a naive approach presented in #420:
> - Find an additional `<tuplet type="stop">` without pending `<tuplet type="start">` registered.
> - In that case, iterate backward over previous notes of the *same voice*, as long as they have the same `<time-> modification>` as the "end" note, and they do not have any `<tuplet>` attribute.
>   - Should we search only on the current measure? Tuplets crossing bar lines are extremely rare;
> - Set the `start_note` of the tuplet to the last note verifying the previous conditions.
> 
> In case of missing `<tuplet type="stop">`, the same could be applied but iterating forward from the `<tuplet > type="start">` rather than backward. However, detecting missing stops might be a bit more difficult. Indeed, in the case of a missing start, you detect it right away when you encounter a stop without having found a start before, but for missing stops, you need some rules that say "the tuplet should be ended by now, there must be a `stop` missing".

However, there might be more complex cases that wouldn't be covered by this approach (I think of nested tuplets, but there might be other weird cases).

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -1566,6 +1566,19 @@ def handle_tuplets(notations, ongoing, note):
         assert (
             start_tuplet.start_note.start.t <= stop_tuplet.end_note.start.t
         ), "Tuplet start time is after tuplet stop time"
+
+    # check that tuplets start and end notes belong to the same voice
+    for tuplet in starting_tuplets + stopping_tuplets:
+        if (
+            tuplet.start_note is not None
+            and tuplet.end_note is not None
+            and (tuplet.start_note.voice != tuplet.end_note.voice)
+        ):
+            warnings.warn(
+                f"Tuplet start and end notes do not belong to the same voice "
+                f"({tuplet.start_note.voice} != {tuplet.end_note.voice}). This might "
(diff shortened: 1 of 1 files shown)


manoskary (approved):
This PR addresses potential errors in tuplets by raising a warning, it does not propose a solution or re-assignment but rather notifies the user about potentially problematic scores.

PR #427, opened 2025-02-04
adjust timing of note events according to tempo changes

huispaty:


partitura/io/importmidi.py
```diff
@@ -287,18 +284,42 @@ def load_performance_midi(
...
+    
+    # adjust timing of events based on tempo changes
+    for pp in pps:
+        for note in pp.notes:
+            note["note_on"] = adjust_time(note["note_on_tick"], tempo_changes, ppq)
+            note["note_off"] = adjust_time(note["note_off_tick"], tempo_changes, ppq)
+        for control in pp.controls:
+            control["time"] = adjust_time(control["time_tick"], tempo_changes, ppq)
+        for program in pp.programs:
+            program["time"] = adjust_time(program["time_tick"], tempo_changes, ppq)
```
CarlosCancino-Chacon, line 301: Should we also adjust the time of `time_signatures`, `key_signatures` and `meta_other`?

partitura/io/importmidi.py
```diff
@@ -287,18 +284,42 @@ def load_performance_midi(
...
 
+def adjust_time(tick, tempo_changes, ppq):
+    """adjust the time of an event based on tempo changes."""
+    time = 0
+    last_tick = 0
+    last_mpq = tempo_changes[0][1]
+    for change_tick, mpq in tempo_changes:
+        if tick < change_tick:
+            break
+        time += (change_tick - last_tick) * (last_mpq / (ppq * 10**6))
```
CarlosCancino-Chacon, line 317: Nice! Maybe we can use the `midi_ticks_to_seconds` (from `partitura.utils.music`) method here, and also below, so that it is clear what this line is computing?. It would be something like

`time += midi_ticks_to_seconds(midi_ticks=(change_tick - last tick), mpq=last_mpq, ppq=ppq)`

(This would be mostly to make clear what this method is computing, but if you think that it makes things less clear, feel free to leave it as it is)

CarlosCancino-Chacon (changes requested):
Very nice and swift bug fix! There are two points that could be addressed:

1. Also adjusting the time of time and key signature messages, as well as meta messages
2. For clarity, use the helper method `midi_ticks_to_seconds` (to make clear what the line is computed
3. Adding typing annotations to `adjust_time`

partitura/io/importmidi.py
huispaty: Thanks for the review! I added all the changes accordingly and extended the test case to also check for changing time and key signatures.

CarlosCancino-Chacon (approved):

PR #428, opened 2025-02-05
Allow `iter_all` to iterate over multiple classes

### LINKED ISSUE #426: Allow `cls` parameter of `iter_all` to take an iterable of classes rather than a single class
When iterating over elements in a part, I sometimes need to iterate over multiple types of elements (e.g. `GenericNote` and `Tuplet`). However, currently, the `iter_all` method only allows searching for one class (eventually including subclasses).

If I want to get all `GenericNote` and `Tuplet` objects, I can currently think of two workarounds:
- `iter_all` over all elements, and then filtering only for `GenericNote` and `Tuplet` objects
  - Advantage: very flexible
  - Disadvantage: very slow (as we iterate over all objects instead of just the ones we are interested in)
- `iter_all` once for each `cls` I need and merge the two results somehow
  - Advantage: iteration faster as we only iterate over `GenericNote` the first time, and `Tuplet` the second time (rather than over all objects as before)
  - Disadvantages:
    - Still a bit slow as iterating twice over all timepoints, once for `GenericNote` and once for `Tuplet` (and it would be even slower if I would like to gather 3, 4... types of objects)
    - Needs merging the results somehow, which might not be directly straightforward with iterators, as we need to compare the start time of the elements coming for each iterator in order to have the final iterator still sorted by time

I feel like an interface like the following could be quite handy:
```py
part.iter_all(cls=[GenericNote, Tuplet], include_subclasses=True)
```
This would return an iterator with all `GenericNote` and `Tuplet` object (and their subclasses). In case of simultaneous event of both types, the order would be the same as the order of classes passed to the function, i.e. in this case all `GenericNote` first, then all `Tuplet`. Note that this iterator would only iterate once over all timepoints, and only on required classes, which would make it faster than the two workaround explained above.

Note that the `include_subclasses` parameter (as well as other parameters) would apply to all classes. I might be possible to allow passing a list of values (of the same size as `cls`) specifying the parameter for each class, but I'm not sure that it would be that useful. Let me know if you think it would.

I'm going to have a go at implementing that soon. I still have one interrogation: what if the user passes `cls=[GenericNote, Note]` with `include_subclasses=True`? (or even `cls=[Note, Note]`) Should the `Note` objects be duplicated? Or should each object only appear once anyway? Or should the function raise an exception? I would be interested in hearing your thoughts about that.

leleogere:
Fixes #426.

When iterating over a `Part`'s elements, the user can only provide a single class to filter the output. If they want to get multiple classes, they can either iterate over *all* classes and filter a posteriori, or iterate multiple times over a single class. Those two workarounds are quite slow, so the possibility to iterate over multiple classes in a single pass would be useful.

This PR simply allows the user to pass an iterable to the `cls` parameter of `iter_all`. For example, the following would allow to easily retrieve a list of all key/time signature/clef change over the score (in that order when they are simultaneous), in a single iteration over all time points.
```py
part.iter_all(cls=[TimeSignature, KeySignature, Clef])
```

There are still a couple of things to discuss before merging this, briefly evoked in #426:
- It makes sense that the parameters `start`, `end` and `mode` apply to all classes. However, should the parameter `include_subclasses` apply to all of them? Maybe I could add the possibility to pass either a single boolean (apply to all classes) or an iterable of booleans (apply to indivisual classes, same size as the `cls` iterable).
- What if a user provides `cls=[GenericNote, Note]` with `include_subclasses=True`? (Or even simpler if they provide `cls=[Note, Note]`) Should the iterator return duplicate objects? Or deduplicated objects only? Or raise an error?

I have not added any tests yet, but I can add some, although I'm not sure in which file they would belong. I haven't found any existing test for the function `iter_all`.

### DIFF
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -1090,8 +1090,8 @@ class Part(object):
 
         Parameters
         ----------
-        cls : class, optional
-            The class of objects to iterate over. If omitted, iterate
+        cls : class or iterable of classes, optional
+            The class (or classes) of objects to iterate over. If omitted, iterate
             over all objects in the part.
         start : :class:`TimePoint`, optional
             The start of the interval to search. If omitted or None,
@@ -1135,12 +1135,21 @@ class Part(object):
             cls = object
             include_subclasses = True
 
+        # If single class passed, convert to list
+        if isinstance(cls, type):
+            cls = [cls]
+
+        if not isinstance(cls, Iterable):
+            raise TypeError("cls should be a class, an iterable of classes, or None")
+
         if mode == "ending":
             for tp in self._points[start_idx:end_idx]:
-                yield from tp.iter_ending(cls, include_subclasses)
+                for c in cls:
+                    yield from tp.iter_ending(c, include_subclasses)
         else:
             for tp in self._points[start_idx:end_idx]:
-                yield from tp.iter_starting(cls, include_subclasses)
+                for c in cls:
+                    yield from tp.iter_starting(c, include_subclasses)
 
     def apply(self):
         """Apply all changes to the timeline for objects like octave Shift."""


manoskary:
Hello @leleogere that is a very nice and cool addition to the package. May I ask you to add a test for that, it does not have to be much, you can use `partitura.EXAMPLE_MUSICXML` and iter through that, no need to create a new test file you could add it to an existing one. Thanks!!!

manoskary:
Hello @leleogere thanks for your message. Let me reply here.
If feels like all proposed solutions are valid but I think in this case, choosing the most straightforward solution for the user would be best in my opinion. I would leave the `include_subclasses` as a `bool` instead of a list, to facilitate user based implementation. If you require such an action, where you would like to specify for each class individually options for it, then probably individual iters would be the solution.

For the case where `cls=[GenericNote, Note] ` with `include_subclasses=True` or `cls=[Note, Note]` we should not return duplicates. Then the decision is in your hands, you could either raise an error or curate the user's input and raise a warning.


But that is just my opinion on this topic.

leleogere:
I move the class iteration logic from `iter_all` to `iter_starting` and `iter_ending`. That way, those methods will benefit of the feature too (and no need to update `iter_all` anymore as it is based on them). I have not taken the time to write the deduplication logic, as well as the tests yet, I'll turn this PR to draft until I find the time to do that.

manoskary:
@leleogere Recently there has been a discussion about `iter_all`. Is this PR also affected, or should I move to review?

mgrachten:
Based on the changes in https://github.com/CPJKU/partitura/pull/451, I created a new PR for this functionality: https://github.com/CPJKU/partitura/pull/457.
This draft can be closed once #457 is merged.

PR #429, opened 2025-02-05
Support for `<tuplet-dot/>` and `<normal-dot/>` tags when parsing tuplets

leleogere:
Following my previous PR #414, I noticed that it didn't cover all possible tuplets cases. In particular, in addition to allow things like "5 eighth notes in the space of 2 quarter notes", MusicXML also allows things like "5 eighth notes in the space of 2 **dotted** quarter notes". My previous implementation wasn't taking into account the eventual dots in the tuplet normal duration.

Those dots can be seen in two places:
- As `<tuplet-dot/>` in `<tuplet-normal>` or `<tuplet-actual>` tags (themselves in `<tuplet type="start">` tags)
- As `<normal-dot/>` tags in `<time-modification>` tags (not sure why there isn't any `<actual-dot/>` in the MusicXML spec)

I've adapted the tuplet test to test one of those dotted cases too.

References:
- https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/tuplet-normal
- https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/tuplet-actual
- https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/time-modification/

### DIFF
--- a/partitura/io/exportmusicxml.py
+++ b/partitura/io/exportmusicxml.py
@@ -242,6 +242,8 @@ def make_note_el(note, dur, voice, counter, n_of_staves):
             tuplet_actual_notes_e.text = str(tuplet.actual_notes)
             tuplet_actual_type_e = etree.SubElement(tuplet_actual_e, "tuplet-type")
             tuplet_actual_type_e.text = str(tuplet.actual_type)
+            for _ in range(tuplet.actual_dots):
+                etree.SubElement(tuplet_actual_e, "tuplet-dot")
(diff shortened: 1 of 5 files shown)


manoskary (approved):
This PR addresses additional tuplet attributes such as tuplet-dot and normal-dots that are sometimes present in musicxml files. It contributes import/export functionalities for musicxml and enriches the Tupler class in score. It also includes more extensive testing for such cases.

PR #441, opened 2025-03-10
fixed performance notes negative tick durations bug

huispaty:
The fix modifies `note_off_tick` and `duration_tick` in a performance note array (in a somewhat inelegant way due to the fixed and default `pps`/`pqs`). If the notes used to generate the note array have an (adjusted) `tick_off` value, that value is used; otherwise, the previous approach is maintained.

### DIFF
--- a/partitura/performance.py
+++ b/partitura/performance.py
@@ -135,7 +135,7 @@ class PerformedPart(object):
         """
         self._sustain_pedal_threshold = value
         if len(self.notes) > 0:
-            adjust_offsets_w_sustain(
+            adjust_note_offsets_with_sustain(
                 self.notes, self.controls, self._sustain_pedal_threshold
             )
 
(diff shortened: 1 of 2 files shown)


partitura/performance.py
```diff
@@ -170,19 +171,28 @@ def note_array(self, *args, **kwargs) -> np.ndarray:
...
                 )
-                - note_on_tick
-            )
+                duration_tick = n.get(
+                    seconds_to_midi_ticks(n["sound_off"], mpq=self.mpq, ppq=self.ppq),
+                    seconds_to_midi_ticks(n["note_off"], mpq=self.mpq, ppq=self.ppq)
+                ) - note_on_tick
+            else:
+                note_off_tick = n["note_off_tick"]
+                duration_tick = note_off_tick - note_on_tick
```
CarlosCancino-Chacon, line 194: Do you think it might be worthwhile to add another "layer" to avoid negative durations by e.g., clipping negative durations at this point (e.g., `duration_tick = max(duration_tick, 0)`). On the other hand this might introduce some issues that might be hard to debug.

CarlosCancino-Chacon (changes requested):
Thanks for the bug fix! I have just a minor comment regarding raising a deprecation warning for renaming one of the functions (for compatibility reasons), and a couple of questions.

PR #442, opened 2025-03-18
Accept int64 values for 'onset_div' in note array

C-Suhit:
Convert the 'onset_div' field in note arrays to accept 64 bit integer values in order to accommodate large scores


CarlosCancino-Chacon (changes requested):
Perhaps we should also use int 64 for `duration_div`?

C-Suhit:
Agreed! Changed duration_div to accept int64 as well.

CarlosCancino-Chacon (approved):

PR #444, opened 2025-04-07
Fix removing last point

leleogere:
The current implementation of `Part._remove_point` would fail with an index error when required to remove the last point of the score.:
```py
    def _remove_point(self, tp):
        i = np.searchsorted(self._points, tp)
        if self._points[i] == tp:
            self._points = np.delete(self._points, i)
            if i > 0:
                self._points[i - 1].next = self._points[i]
                self._points[i].prev = self._points[i - 1]
            if i < len(self._points) - 1:
                self._points[i].next = self._points[i + 1]
                self._points[i + 1].prev = self._points[i]
```
When is the last point of the array, the `if i > 0` branch tries to access `self._points[i]` which has been deleted, resulting in an index error.

The new implementation fixes that, and adds tests for that behaviour.

### DIFF
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -950,13 +950,22 @@ class Part(object):
     def _remove_point(self, tp):
         i = np.searchsorted(self._points, tp)
         if self._points[i] == tp:
+            # If not on bound, link prev and next points together
+            if i > 0 and i < len(self._points) - 1:
+                self._points[i - 1].next = self._points[i + 1]
+                self._points[i + 1].prev = self._points[i - 1]
+            # If last point, set previous point as last one
+            elif i > 0:
+                self._points[i - 1].next = None
+            # If first point, set next point as first one
+            elif i < len(self._points) - 1:
+                self._points[i - 1].prev = None
+            # That last case only happen when there is a single point in the array,
+            # no prev/next update in that cas, simply delete
+            else:
+                pass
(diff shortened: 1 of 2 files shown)


mgrachten (approved):
Looks good to me.

`self._points[i]` could still raise an IndexError when called with arbitrary time points, but since this is an internal method that is only called with time points from the time line, I guess we don't need handle that case.

leleogere:
Oh yes I did not think of this case! Even if it's internal, should I add a safety test that would trigger a more meaningful error?

mgrachten:
yes, perhaps simply
```python
if i < len(self._points) and self._points[i] == tp:
```

leleogere:
> yes, perhaps simply
> 
> ```python
> if i < len(self._points) and self._points[i] == tp:
> ```

Do you think trying to remove a non-existing point should do nothing? I think I would prefer to raise an error (either a `ValueError` or the same `IndexError` but with a more meaningful message) as I don't feel like it is supposed to happen, and it could hide another issue if not raised. What do you think about it?

mgrachten:
> Do you think trying to remove a non-existing point should do nothing? 

You could frame the semantics of the non-raising version as "remove `tp` from the timeline if it's there", but I'm not advocating for this approach, and raising a ValueError is perfectly fine (perhaps better).

Issue #443, opened 2025-03-26
Transposition issues: breaks ties, does not account for key signature, inefficient...


leleogere:
Currently, the `transpose` function has several issues, making it difficult to use in practice.

# Issues with `transpose`

## It breaks ties

Currently, the function iterates over `part.note_tied`, which only returns the first note of each tied group. Therefore, only that note is updated, and the subsequent tied note(s) are left untouched, breaking the tie, and resulting them being in the wrong key. 

You can see that in the following example. The first image is the original score, with three ties, and the second one is the same score transposed up by a major third.

![Original score](https://github.com/user-attachments/assets/dc7108fe-2761-4aa9-b564-45ca64d144c4)

![Score transposed up by a major third, with ties broken](https://github.com/user-attachments/assets/8a7ccd90-ddc9-43bb-86fc-c853a0fc3615)

The `transpose` function should either
- iterate over `part.notes` rather than `part.tied_notes` and update all notes
- iterate over `part.tied_notes`, but update all `tie_next` notes when updating each note

I'll likely open a PR for this, feel free to tell me if you prefer one solution over the other.

## It does not update key signatures

When transposing a score, I usually want not only to transpose the notes, but also the key signature. In the previous example, which is in the key of C, I would expect the score transposed by a major third to be in the key of E, therefore having 4 permenent sharps at the clef.

This could also be a boolean argument of the `transpose` function like `update_key_signatures`, being `False` by default to preserver old behaviour.

## It breaks for large scores

Sometimes, when transposing large scores, I encounter recursion errors. They are due to the deep copy that happens before the actual transposition, which is done in place.

This must have been a problem before because currently, the recursion limit is manually and arbitrarily increased to 10000. However, this has one major issue: if 10000 is not enough, they there is no way for the user to make the function work except by modifying the source files (usually when I encounter recursion errors, I increase the limit in my code and it works, but not in that case as Partitura is modifying it itself).

https://github.com/CPJKU/partitura/blob/bfaac43eebb47eaf1aef9841087674c08b22a30e/partitura/utils/music.py#L300-L305

I can see two solutions:
- Pass the recursion limit as parameter (does not feel right to me)
- Remove those lines and let the user manage themselves the recursion limit (eventually adding a warning in the doc that one might have to increase the limit).

## It is not efficient

If I pass a `Score` object to the function, it will deep copy it, then iterate over all parts, deep copying them too, which is not needed.

[...]

PR #449, opened 2025-04-24
Fix transposition issues

leleogere:
*Note: this was first pushed to develop by mistake, I reverted the commit.*

In #443, I raised a couple of concerns regarding the transposition implementation.

In this PR, I'm trying to adress them.

## Breaking ties

Transposing does not break ties anymore. Previously, the only notes that were transposed were `part.notes_tied`, so any subsequent tied notes would be left unmodified.

In my implementation, I added a boolean parameter `update_ties` to `_transpose_note_inplace`, and kept the loop over `part.notes_tied`. Alternatively, I could simply loop over `part.notes` and drop this `update_ties` parameter. What do you think?

## Update key signature

The former implementation was only transposing notes, without affecting the key signature. I've added a parameter `transpose_key_signatures` to `.transpose`, left to `False` by default to keep the old behaviour, that will also transpose the key signatures into the new key. Do you prefer it `True` by default?

I've added a test for this behaviour.

## Handle large scores

Previously, the transpose function was systematically creating a deep copy of the score (and sometimes multiple), which can be very ineffective. Additionally, the recursion limit was manually increased with a hardcoded value, which was preventing to pass scores that would need larger limits.

To address that, I added an `inplace` parameter, that will speed things up if the user wants to directly modify its score. I also removed the recursion limit, and added a sentence in the doc explaining that the user might want to explicitly raise that limit when dealing with large scores and `inplace=False`.

This parameter also allows to fix some bugs in deep copies management that were present before (see original issue for more insights).

## Fix transposing down

In my tests, I discovered that while transposing up was working ok, transposing down was broken, and lead to some quadruple-flats or even septuple-sharps!

I fixed the implementation so that we can now pass ascending or descending intervals without issue. I adapted the tests to test descending interval.

## General refactoring and breaking change

When dealing with this PR, I came across some parts that were very similar in the function `transpose_note` and `_transpose_note_inplace`. The former was dealing with `step` (str) and `alter` (int) (no octave), and the second one were dealing with a `Note` object (including step, alteration and octave). In order to limit duplication, I moved all the transposition logic into a `transpose_note_attributes` function, that deals with step, alteration and octave altogether.

As the name suggest, this function does not operate on a `Note` object, but on individual attributes (like the former `transpose_note`). It takes an `Interval` object, a step (str), an alteration(int) and an octave (int). It has default values for alteration (0) and octave (0), so that one can choose not to use them if not needed.

The issue with that is, as I want default values for `alter` and `octave`, I can't have the `interval` parameter last as before. I could have it in second like `transpose_note_attributes(step, interval, alter=0, octave=0)`, but it feels weird to separate `step` for the other attributes. Finally, I decided to go with `transpose_note_attributes(interval, step, alter=0, octave=0)`, but it's not consistent with the order of `pt.music.transpose(score, interval)`. I don't know if you have any preference about this.

In consequence, I removed the `transpose_note` function, that I was finding quite confusing as it was not dealing with `Note` objects, and couldn't deal with octaves. This is a breaking change, maybe I should deprecate it and call `transpose_note_attributes` under the hood? Let me know if you would rather keep it around.

I replaced all `transpose_note` references in the library with the new `transpose_note_attributes`.

### DIFF
--- a/partitura/utils/music.py
+++ b/partitura/utils/music.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
     # For this to work we need to import annotations from __future__
     # Solution from
     # https://medium.com/quick-code/python-type-hinting-eliminating-importerror-due-to-circular-imports-265dfb0580f8
-    from partitura.score import ScoreLike, Interval
+    from partitura.score import ScoreLike, Interval, Note, KeySignature
     from partitura.performance import PerformanceLike, Performance, PerformedPart
 
 
@@ -143,24 +143,24 @@ def ensure_rest_array(restarray_or_part, *args, **kwargs):
         )
 
 
-def _transpose_step(step, interval, direction):
+def _transpose_step(step: str, num_steps: int, direction: str):
     """
     Transpose a note by a given interval.
     Parameters
     ----------
-    step
-    inverval
-
+    step: Step to transpose
+    num_steps: Number of steps (not semitones) to transpose
+    direction: Direction of the transposition
     """
     op = lambda x, y: abs(x + y) % 7 if direction == "up" else abs(x - y) % 7
-    if interval == "P1":
+    if num_steps == 1:
         pass
     else:
-        step = STEPS[op(STEPS[step.capitalize()], interval - 1)]
+        step = STEPS[op(STEPS[step.capitalize()], num_steps - 1)]
     return step
 
 
-def _transpose_note_inplace(note, interval):
+def _transpose_note_inplace(note: Note, interval: Interval, update_ties: bool = True):
     """
     Transpose a note by a given interval.
     Parameters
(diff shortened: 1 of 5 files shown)


manoskary:
@leleogere Thank you for this PR. You are indeed addressing very important issues with transposition
Personally, I agree with your all your points. Regarding the last breaking change, I believe the decision to use this way to transpose notes was going in parallel with needs of packages that work with music21 or generally with note_arrays where effectively you have note_attributes instead of note objects. Effectively, when transposing an entire part or score we would need different tools than when only transposing a `note_array`. As you can imagine now, the functionalities where designed to transpose principally note arrays and that's why it was only focused on tied notes, but effectively that was not a perfect nor complete solution. I would propose to not have breaking changes, but rather add a warning that in some later release the name of the function might change and add also the future function name as you indicated already.

leleogere:
I'm not sure to understand correctly. You would prefer:
- Keeping the `transpose_note(step, alter, interval)` as well as `transpose_note_attributes` functions to avoid the breaking change (then `transpose_note` would call the new `transpose_note_attributes` under the hood)
- Not introducing the new `transpose_note_attributes`, keeping `transpose_note` instead and warn in release notes that it will be renamed at some point `transpose_note_attributes` -> not easily feasible as it's more than a simple renaming: while `transpose_note_attributes` aims at replacing and clarifying `transpose_note`, it also introduces a new behaviour (octave management) and change the function signature
- Keeping `transpose_note` around for now but marking it as deprecated in the release notes as well as raising a deprecation warning when used, so that people have time to adapt their code
- Something else?

partitura/utils/music.py
```diff
@@ -172,82 +172,58 @@ def _transpose_note_inplace(note, interval):
...
+    ks.fifths = new_fifths
 
 
-def transpose_note(step, alter, interval):
+def transpose_note_attributes(
+    interval: Interval,
+    step: str,
+    alter: int = 0,
+    octave: int = 0,
+):
```
manoskary, line 209: I recommend not to deprecate the `transpose_note` function. You can add a warning or future deprecation flag as a decorator on top of the function `@deprecated("Starting with version X sse transpose_note_attributes instead")`

manoskary (changes requested):
This PR refactors and improves the note transposition logic in partitura. It also adds key and in place transposition to avoid copying parts. Only very minor changes are requested.

- Introduces a new function, `transpose_note_attributes`, for flexible note transposition (handles step, alter, octave, direction) to replace the previous `transpose_note` function
- Adds support for transposing key signatures.
- Refactors codebase to use the new function and improves robustness including better documentation and typing.
- Updates and extends tests to verify note and key signature transposition.
- Updates the example MusicXML to include a key signature.

leleogere:
I reintroduced `transpose_note` as a deprecated function, arbitrarily choosing 1.8.0 for the removal. Let me know if you would prefer a latter version.

I also noticed that the method `_transpose_step` was not used anymore with my changes, so I removed it. As it is supposed to be a private function (starting with `_`), I did not mark it as deprecated first as it's not supposed to be used by users. If you prefer making it as deprecated first anyway, I can do it.

manoskary (approved):
Last changes implement proposed discussion. This PR is a nice addition making the slightly problematic note transposition functions more robust.

Issue #446, opened 2025-04-23
speed of iter_all


rbdannenberg:
I'm measuring iter_all() at over 20ms per note when applied to a MIDI file., e.g.
```
    timer.start()
    for item in ppart.iter_all():
        pass
    timer.stop(report=True)
```
 Is that normal?

CarlosCancino-Chacon:
Hi @rbdannenberg! Thanks for opening this issue.  This is indeed quite slow. I did a quick test on piano scores and I get an average of around 2ms

![Image](https://github.com/user-attachments/assets/9637903d-b4ce-4d9c-ae9d-c9c2d2456ee2)

What is the context in which you are using  `iter_all`? If you are iterating over the notes, explicitly iterating over `Note` elements is a lot faster:

```python
for note in ppart.iter_all(partitura.score.Note):
    pass
```

![Image](https://github.com/user-attachments/assets/5ad0d86a-d944-4dc7-8ecf-f6e855f2cc4b)

Otherwise, could you perhaps send the file(s) where you encounter this issue?

rbdannenberg:
I was testing with this file: [sarabande.mid](https://github.com/music-computing/amads/blob/main/amads/music/midi/sarabande.mid).
We're using partitura to import music xml and midi files to a different (not partitura) representation, so I'm doing iter_all() to access everything.
Some things I find puzzling:
- I believe Python's virtual machine instruction time is (order of magnitude) 100ns, so 0.0013s to access the next note corresponds to 13,000 instructions. That seems way too high.
- I measured about 0.19s to import the 301 notes from MIDI file and construct Score in Partitura, which is about 0.00063s/note. So it is much faster, based on your measurements, to construct the entire score than to access even just the notes in the score.

rbdannenberg:
In score.py, the for loop in this code:
```
        if include_subclasses:
            for subcls in iter_subclasses(cls):
                global iter_subclasses_count
                iter_subclasses_count += 1
                yield from self.starting_objects[subcls]
```
is executed 28482 times on a score with 301 notes and relatively few other objects, which means about 95 times per note.
I think that explains the order of magnitude of the cost of iter_all(). It looks like maybe the classes of interest are inheriting from a large class hierarchy outside of the Partitura types.

mgrachten:
> It looks like maybe the classes of interest are inheriting from a large class hierarchy outside of the Partitura types.

[...]

PR #451, opened 2025-05-19
Speedup `iter_all`

mgrachten:
Using `Part.iter_all()` without specifying `cls` sets the class to `object`, and sets `iter_subclasses = True`. This has the unintended consequence that the full python class hierarchy is traversed at each time step. Setting `cls = TimedObject` is equivalent (since any starting/ending objects should be subclasses of TimedObject), and greatly reduces the number of classes to be iterated over.

In my minimal tests, this change gives approx. a 100x speedup.

Fixes https://github.com/CPJKU/partitura/issues/446

p.s. There may be more opportunity for optimization. I will look at this later.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Release Notes
 =============
 
+
+### Other Changes
+- speedup Part.iter_all
+
 Version 1.6.0 (Released on 2025-02-27)
 --------------------------------------
 


partitura/score.py
```diff
@@ -28,7 +28,6 @@
 from partitura.utils import (
     ComparableMixin,
     ReplaceRefMixin,
-    iter_subclasses,
```
leleogere, line 31: Is `iter_subclasses` used somewhere else in the library? If not, I would suggest removing it completely

partitura/score.py
```diff
@@ -1477,10 +1476,12 @@ def iter_starting(self, cls, include_subclasses=False):
...
         """
-        yield from self.starting_objects[cls]
         if include_subclasses:
-            for subcls in iter_subclasses(cls):
-                yield from self.starting_objects[subcls]
+            for key, items in self.starting_objects.items():
+                if issubclass(key, cls):
+                    yield from items
+        elif cls in self.starting_objects:
+            yield from self.starting_objects[cls]
```
leleogere, line 1484: I am not familiar with the working of iter_subclasses, but does this change have an impact on the order the objects are yield by iter_all? If so, this would technically be a breaking change, however I don't think that the order is documented anywhere, so I guess it might be OK.

leleogere (commented):
Thank you for this optimization! It makes much more sense to iterate only over partitura objects, rather than all python hierarchy. I just left a couple of comments/questions.

partitura/score.py
mgrachten: Good point. I think the order of objects is likely to change. But as you state, the order of the objects is an undocumented implementation detail, so that shouldn't really keep us from merging this. A warning in the changelog would be prudent though.

partitura/score.py
mgrachten: It's not being used anywhere in the package. I also think we should remove it. As with the comment below, the function is not referenced in the official documentation.

CarlosCancino-Chacon (approved):
Thanks @mgrachten!

PR #459, opened 2025-06-10
Round ticks in `save_score_midi`

ilya16:
Hi,

This PR fixes a bug that causes incorrect MIDI ticks in some scores saved as MIDI files.

I have attached an archive containing a `score.mxl` file on which the issue persists. The statement `ppq * (qm(t) - ftp)` sometimes produces numbers like `x.9999999`, which are truncated to `x` rather than being rounded to `x + 1` as expected. Consequently, some notes are shorter or longer than in the original score.

The same issue with round might probably be relevant to other parts of the package. However, I haven't checked all the `int()` calls.

[example.zip](https://github.com/user-attachments/files/20674167/example.zip)

image

### DIFF
--- a/partitura/io/exportmidi.py
+++ b/partitura/io/exportmidi.py
@@ -395,7 +395,7 @@ def save_score_midi(
 
         def to_ppq(t):
             # convert div times to new ppq
-            return int(ppq * (qm(t) - ftp))
+            return round(ppq * (qm(t) - ftp))
 
         for tp in part.iter_all(score.Tempo):
             tempos[to_ppq(tp.start.t)] = MetaMessage(


leleogere (approved):
This PR fixes a rounding bug in MIDI export, LGTM.

leleogere:
> I would like to request a small change: could you perhaps cast the result as an integer (i.e., `return int(round(ppq * (qm(t) - ftp))`? That way the function will return the same data type as before.

If no digits are specified (`ndigits=None`), the round function does already return an `int`, doesn't it?

CarlosCancino-Chacon:
Hi @leleogere!
 
> If no digits are specified (`ndigits=None`), the round function does already return an `int`, doesn't it?

Not necessarily. For example `type(np.round(5.99999)` will return an`np.float64`  if `x` is a `float`.  It will return an `int` (or a numpy integer type if the argument to round is an integer to begin with.

CarlosCancino-Chacon:
@ilya16 and @leleogere! Thanks for the clarification. Somehow I was convinced that `round` and `np.round` worked the same. I approved the PR

PR #461, opened 2025-06-27
Fix some reference issues when creating a ScoreVariant

### LINKED ISSUE #450: Recursion error when pickling an unfolded score
It seems that unfolding repetitions in a score results in the unfolded score not being picklable because of some recursion error.

```python
from pathlib import Path
import pickle
import sys

import partitura as pt
import partitura.score


sys.setrecursionlimit(2**31 - 1)  # Maximum value allowed by python


score_path = Path("~/Documents/datasets/asap-dataset/Bach/Fugue/bwv_854/xml_score.musicxml").expanduser()
dump_file = Path("/tmp/test.pkl")

score = pt.load_musicxml(score_path)[0]
with dump_file.open("wb") as f:
    pickle.dump(score, f)
print("Original score dumped")

unfolded_score = pt.score.unfold_part_maximal(score)
with dump_file.open("wb") as f:
    pickle.dump(unfolded_score, f)  # Triggers recursion error
print("Unfolded score dumped")
```
```python-traceback
Original score dumped
Traceback (most recent call last):
  File "/home/gerel/.config/JetBrains/PyCharm2025.1/scratches/scratch_85.py", line 26, in <module>
    pickle.dump(unfolded_score, f)  # Triggers recursion error
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded while pickling an object
```

This happens for scores with repetitions, but also for scores *without* repetitions (in which case the `unfold_part_maximal` should not have any effect). I guess something is broken in the references update.

leleogere:
Closes #450.

This PR adds a couple of attributes that were not correctly handled when a `ScoreVariant` would be produced, resulting in `RecursionError`s (and even segmentation faults with some scores) when trying to pickle some unfolded scores:
```py
import warnings
from pathlib import Path
import pickle
import sys
import partitura as pt
import partitura.score

sys.setrecursionlimit(2**31-1)

score_path = Path("~/Documents/datasets/asap-dataset/Liszt/Sonata/xml_score.musicxml").expanduser()

with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    score = pt.load_musicxml(score_path)
    score = pt.score.unfold_part_maximal(score)
with open("/tmp/test.pkl", "wb") as f:
    pickle.dump(score, f)
# Segmentation fault (core dumped)
```

The added references are the following:
- `GenericNote.fermata`
- `Note.beam`
- `UnpitchedNote.beam`
- `Beam.notes`
- `Fermata.ref`  **EDIT: Seems problematic, see comment below**

Before, those attributes were not updated during the creation of a `ScoreVariant`, resulting in them pointing to the elements in the original score. Now, the references are correctly updated to point to the corresponding elements in the created `ScoreVariant`.

### DIFF
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -1845,13 +1845,13 @@ class GenericNote(TimedObject):
         # maintain a list of attributes to update when cloning this instance
         self._ref_attrs.extend(
             [
+                "fermata",
                 "tie_prev",
                 "tie_next",
                 "slur_stops",
                 "slur_starts",
                 "tuplet_stops",
                 "tuplet_starts",
-                "fermata",
             ]
         )
 
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -2194,6 +2194,8 @@ class UnpitchedNote(GenericNote):
         if self.beam is not None:
             self.beam.append(self)
 
+        self._ref_attrs.append("beam")
+
     def __str__(self):
         return " ".join(
             (
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -1851,6 +1851,7 @@ class GenericNote(TimedObject):
                 "slur_starts",
                 "tuplet_stops",
                 "tuplet_starts",
+                "fermata",
             ]
         )
 
@@ -2117,6 +2118,7 @@ class Note(GenericNote):
         self.octave = octave
         self.alter = alter
         self.beam = None
+        self._ref_attrs.append("beam")
 
     def assign_beam(self, beam):
         self.beam = beam
@@ -2228,6 +2230,7 @@ class Beam(TimedObject):
         super().__init__()
         self.id = id
         self.notes = []
+        self._ref_attrs.append("notes")
 
     def append(self, note):
         note.beam = self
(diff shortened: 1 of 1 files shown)


leleogere:
> Actually, I don't understand why it gets the location attribute a second time, when it's already defined above. It seems like repeated tweaks to this code over time made it somewhat inconsistent.

It seems that there might be quite a few problems in that barline code indeed. A few things that I find strange are:
- the `Barline` is added on the timeline at `position`, shouldn't it be added at `position_barline` (taking into account the `location` attribute)?
- the `Barline` is only added if the `bar-style` attribute is not `None`, which mean that we wouldn't be able to set the `ref` attribute of a fermata on a a barline without `bar-style`.
- If the barline location is `None`, and it has a fermata, the fermata location position is assumed to be `"right"` according to the comment, and appended to the `trailing_children` list containing elements that should be added at the end of the measure. However, if the location is explicitly `"right"`, the fermata is directly added at the current position in the timeline, rather than appended in `trailing_children`.

There might be other issues, but that's the first ones that came to my mind when reading this code.

mgrachten:
All valid points. Since you have a good grasp on the issues, feel free to resolve as you see fit. I currently don't have the bandwidth to address this.

sildater:
keep open

Issue #462, opened 2025-07-10
Error in loading compressed .mxl files — suggested fix


mimbres:
### Tested environments
- Ubuntu 23.04 LTS / macOS 15.5
- Python 3.10
- Partitura 1.6.0

### Problem 😤
https://github.com/CPJKU/partitura/blob/bfaac43eebb47eaf1aef9841087674c08b22a30e/partitura/io/importmusicxml.py#L236

When loading a compressed `.mxl` file, `Partitura` updated for #65  selects the last file (`[-1]`) from the unzipped contents. However, this can lead to incorrect behavior.

```python
print(zipped_xml.namelist())
# Output:
# ['musicXML.xml', 'META-INF/', 'META-INF/container.xml']
```
In this case, `musicXML.xml` is the actual score file, but [-1] refers to 'META-INF/container.xml', which is not the intended target.

### Fix
A simple fix would be to select [0] instead.
```python
    if isinstance(filename, str):
        if zipfile.is_zipfile(filename):
            with zipfile.ZipFile(filename) as zipped_xml:
                contained_xml_name = zipped_xml.namelist()[0] # <-- not [-1] !!
                xml = zipped_xml.open(contained_xml_name)
```

 While this works in my environment, it might be OS- or Software- (e.g. musescore) dependent, so I'm raising an issue instead of submitting a PR for now.

Let me know what you think!

leleogere:
I'm working on another fix on mxl reading, and I will incorporate a fix for that too. A more robust solution would be to follow the [MusicXML specification](https://www.w3.org/2021/06/musicxml40/tutorial/compressed-mxl-files/):

> As an example, let us look at the Dichterliebe01.mxl file available on the MakeMusic web site. Here is its META-INF/container.xml file:
> 
> ```
> <?xml version="1.0" encoding="UTF-8">
> <container>
>   <rootfiles>
>     <rootfile full-path="Dichterliebe01.musicxml"
>               media-type="application/vnd.recordare.musicxml+xml"/>
>   </rootfiles>
> </container>
> ```
> 
[...]

PR #467, opened 2025-09-03
Allow parsing mxl when using Pathlib

leleogere:
Partitura MusicXML import currently only allows decompressing mxl files when the path is provided as a string. This PR simply triggers the decompression mechanism also when a pathlib object is passed.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -5,6 +5,7 @@ This module contains methods for importing MusicXML files.
 """
 
 import os
+from pathlib import Path
 import warnings
 import zipfile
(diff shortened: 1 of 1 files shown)


manoskary (approved):
This PR adds an additional type when instance checking paths for decompressing mxl to xml files.

CarlosCancino-Chacon:
Hi @leleogere and @manoskary! Should I merge this PR into develop, or would you create a new PR for addressing #462?

CarlosCancino-Chacon:
In that case, I think it would be easier to open a new PR for clarity. I will merge this PR.

PR #468, opened 2025-09-03
Follow MusicXML guidelines to parse `mxl` compressed files

### LINKED ISSUE #462: Error in loading compressed .mxl files — suggested fix

leleogere:
Fixes #462.

Currently, in a compressed MusicXML file (`mxl`), Partitura assumes that the underlying XML file containing the score data is the last XML file of the zip archive. However, #462 suggests that different systems (?), might list files in zip archives differently, leading to errors in parsing.

A more robust way to deal with that is to follow the [MusicXML specification](https://www.w3.org/2021/06/musicxml40/tutorial/compressed-mxl-files/):
- In the zip archive, open the file `META-INF/container.xml`
- Find the first `<rootfile/>` tag
- Extract the `full-path` attribute, containing the path of the score in the archive
- Parse the score

This avoids relying on the parsing order of the zip archive.


CarlosCancino-Chacon:
Hi @leleogere! Thanks for the nice solution. I added a small test as well to increase the coverage 😉

PR #481, opened 2026-02-02
Refactor installation

sildater:
This PR refactors the `setup.py` based installation to a `pyproject.toml` based one. Optional dependencies are now specified in the main toml file, the requirements are no longer needed, and the testing workflow was adapted to the installation process. I also added a newer python version to the CI tests.

### DIFF
--- a/.github/workflows/partitura_unittests.yml
+++ b/.github/workflows/partitura_unittests.yml
@@ -16,7 +16,7 @@ jobs:
       max-parallel: 5
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.10","3.11","3.12","3.13","3.14"]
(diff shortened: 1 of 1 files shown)


sildater:
ok, so maybe not :) multiple options are available:
- remove the 3.13 installation and test only against 3.8 (which has been end-of-life since 2024)
- adapt the optional installs for the 3.13 test, have two configurations
- remove 3.8, bump our lowest supported version (currently 3.7) to something still used ([3.10 for example](https://devguide.python.org/versions/)) and configure the requirements for this new python

I have no clear favorite, the reviewers can say what should be done.

mgrachten:
> Would you see an issue with this @mgrachten?

To the contrary, I'm all for >=3.10.

PR #495, opened 2026-03-26
Enhancement: Add 'local_grace_order' field to score note array when include_grace_notes

C-Suhit:
There are situations in certain scores such as Rachmaninoff's Piano Concerto No. 3, Movement 1, where we notice multiple grace notes being associated with a specific beat in the score. 
On some occasions, these grace notes have an intended sequential order within them, as seen in the image below:

**Fig. 1:**
Acciaccatura_non_chord

On other occasions, the grace notes are intended to be played together as a chord, as seen in the following image:

**Fig. 2:**
Acciacatura_chord

While this information is provided in the MusicXML score in terms of the document order in which these grace note elements appear, it is currently not passed on to the score note array.

The enhancement associated with this pull request allows for this information to be passed on to the score note array by creating a new field in the array titled 'local_grace_order' when the 'include_grace_notes' flag is True while creating a note array from a score part.
For a given onset unit and a given voice, all the respective grace notes are assigned integers values from 0 upwards according to their document order, while all non-grace notes are assigned the value -1. The grace notes in **Fig. 3** below will therefore be assigned the 'local_grace_order' values 0, 1, 2, 3 and 4 respectively.

**Fig. 3:**

non_chord_example

If such grace notes contain the <chord/> tag within the MusicXML document, all such grace notes are given the same 'local_grace_order' value, as they are intended to play together. All the grace notes in **Fig. 2** will be assigned the 'local_grace_order' value of 0.

In the case of situations where there is a mix of sequential and chordal grace notes, such as in this image:

**Fig. 4:**
Acciaccatura_mixed_chord

the first three grace notes are assigned the 'local_grace_order' values 0, 1 and 2 respectively, and the chord of four grace notes are all assigned the value 3.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -1337,10 +1337,14 @@ def _handle_note(e, position, part, ongoing, prev_note, doc_order, prev_beam=Non
     # code in handle_tuplet function)
 
     chord = e.find("chord")
+    is_grace_chord = False
     if chord is not None:
         # this note starts at the same position as the previous note, and has
         # same duration
         assert prev_note is not None
+        is_grace_chord = True
+        if prev_note.is_grace_chord == False:
+            prev_note.is_grace_chord = True
         position = prev_note.start.t
         duration = prev_note.duration
         duration_from_symbolic = prev_note.duration_from_symbolic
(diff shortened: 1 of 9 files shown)


sildater:
looks good as far as I can tell! thank you for the detailed explanations in the PR! can be merged unless further additions are planned to be part of this PR.

sildater (approved):

PR #496, opened 2026-03-31
Score directions maps

huispaty:
- adds score direction maps for tempo and dynamics directions
- fixes a bug for tempo directions that are abbreviated with a trailing period (e.g., ten. -> tenuto) were not recognized as valid directions

### DIFF
--- a/partitura/score.py
+++ b/partitura/score.py
@@ -799,9 +799,10 @@ class Part(object):
 
     @property
     def tempo_directions_map(self):
-        """A function mapping timeline times to the ConstantTempoDirection
-        active at that time. The function can take scalar values or 
-        lists/arrays of values.
+        """A function mapping timeline times to the TempoDirection
+        active at that time. Returns None for times before the first direction,
+        and fills forward after the last direction.
+        The function can take scalar values or lists/arrays of values.
         
         Returns
         -------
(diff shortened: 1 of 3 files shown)


partitura/score.py
```diff
@@ -711,6 +711,79 @@ def dynamics(self):
...
+        function
+            The mapping function
+        """
+        cd_it = [e for e in self.dynamics if isinstance(e, ConstantLoudnessDirection)]
+
+        if len(cd_it) == 0:
+            warnings.warn("No constant dynamics found, returning None map")
+            return lambda t: np.full(np.asarray(t).shape, None) if np.ndim(t) > 0 else None
+
+        fallback_dur = 2 * self.measures[0].duration
```
CarlosCancino-Chacon, line 737: This fallback duration is always two bars, but wouldn't be updated if there is a change in time signature, right? I guess in principle it is okay to have a default value, even if it is constant, but I wonder whether it might be better to change it to exactly two bars according to the current time signature at the specified points. Also, I don't know if this would create a weird issue for pickup measures.

That being said, it would be okay if this is just documented here somehow, and leave these points to a future PR, if this issue ends up having unexpected consequences.

partitura/score.py
```diff
@@ -724,6 +797,55 @@ def tempo_directions(self):
...
+        function
+            The mapping function
+        """
+        td_it = self.tempo_directions
+        
+        if len(td_it) == 0:
+            warnings.warn("No tempo directions found, returning None map")
+            return lambda t: np.full(np.asarray(t).shape, None) if np.ndim(t) > 0 else None
+        
+        fallback_dur = 2 * self.measures[0].duration
```
CarlosCancino-Chacon, line 822: This is the same point as above. 

The pragmatic solution for now would be to add a comment here and a note in the documentation of the function.

CarlosCancino-Chacon (commented):
Thanks @huispaty for the PR! I would suggest two small changes:

1. I added some comments regarding a potential issue with a default value, but we can address this for the time being by documenting this default duration. 
2. It would be nice to add a test for this method. Attached is a simple MusicXML with dynamics changing every measure that could be used for that.

[example_dynamics.musicxml.zip](https://github.com/user-attachments/files/28063719/example_dynamics.musicxml.zip)

partitura/score.py
huispaty: Hello! 
I added a comment in the function docstring pointing to the fixed fallack duration at this point.
Unfortunately I cannot access your test xml file (the link points to not found) :/ I would suggest a test in the next release together with the functionality change to account for changes in the time signature.

PR #498, opened 2026-04-02
look up ref "sameas" element in MEI parser

sildater:
an element may refer to another using the sameas attribute along with an xml:id and inherit some of the other's properties. this change implements a lookup for the case of durations of rests. the pattern should work in general, but I don't know exactly where else we might encounter it.


C-Suhit (commented):
Looks good! All rows in the score note array has a valid duration_beat value, as expected!

C-Suhit (approved):

PR #502, opened 2026-05-08
refactor tie handling

sildater:
when importing musicxml tie handling happens outside of the note loop to better capture tie starts and stops encoded in arbitrary order in the xml document.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -13,6 +13,7 @@ import zipfile
 from typing import Union, Optional, List
 import numpy as np
 from lxml import etree
+from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


joann8512:
The mentioned issue is fixed. 
Testing with some additional manual additions of ties (multiple ties at one time step), seems like it's also not going across staff -- should we also address it here?

Screen Shot 2026-05-15 at 3 46 08 PM
Screen Shot 2026-05-15 at 3 46 42 PM

sildater:
Thanks for the review! Is this the last onset where the g4 should be tied? That should work, yes. Can you send me yor test file?

joann8512:
Actually... double-checked, and it was an error on my end. Will merge the PR!

joann8512 (approved):

PR #505, opened 2026-05-15
robustness: clearer errors for malformed MusicXML + ppq overflow in save_score_midi

e7mac:
## Summary

Three small fixes turning cryptic crashes into either a warning-and-continue or a clear `ValueError`. All four surface from a single root cause pattern (a missing precondition check at the boundary between user input and an unforgiving downstream library), but they're independent enough that you can take any subset.

Found while running an alignment pipeline against a ~1500-file MusicXML corpus (DCML, kern-derived, Mutopia, ASAP). Empirical impact: clean runs on a stratified 50-file sample went from 38/50 → 43/50.

### 1. `_handle_note`: bare `assert prev_note is not None` on `<chord>` as the first note

```python
chord = e.find("chord")
if chord is not None:
    assert prev_note is not None        # <-- crashes with no message
    position = prev_note.start.t
```

Malformed MusicXML can have `<chord>` on a part's first note (or first after `<backup>`/`<forward>`). In that case there's no previous note to copy timing from, but the import crashes with an empty `AssertionError` deep in the parser, instead of either a useful message or graceful continuation. Replaces with a `warnings.warn(...)` and falls through to treat the note as standalone.

Files this lets load instead of crash: `BachJS_Applicatio.xml`, `HandelGF_Menuet_I.xml`, `ChopinFF_Prélude_22.xml`, `BeethovenLv_Sonata_No__19__*` (all Mutopia).

### 2. `get_ppq`: handle Score with no parts

```python
ppqs = np.concatenate(
    [part.quarter_durations()[:, 1] for part in score.iter_parts(parts)]
)   # ValueError: need at least one array to concatenate
```

When a Score has no Parts (e.g. a MusicXML file with `<part-list/>` and no `<part>` elements — a real pattern in some community-contributed sources), the comprehension is empty and `np.concatenate` crashes. Returns a safe default (480) so the caller can either continue or — better — bail out clearly via fix 3.

### 3. `save_score_midi`: early-raise with a clear message on empty Score

Rather than letting empty-Score errors surface deep inside numpy, raise

```
ValueError: Cannot save to MIDI: the score has no parts. This usually
means the source MusicXML had an empty <part-list/> or no <part> elements.
```

at the top of `save_score_midi`. Users debugging a bad input get an actionable message immediately.

### 4. `get_ppq`: cap ppq at SMF 16-bit header limit (32767) + warn

`np.lcm.reduce(quarter_durations)` over a part with many coprime tuplet denominators can return values well over 32767 (Ravel — Miroirs IV — Alborada del gracioso comes out to ~360000). Mido enforces the SMF type-0/type-1 header constraint and crashes at write time with

```
struct.error: 'h' format requires -32768 <= number <= 32767
```

This caps at 32767 with a `RuntimeWarning` (lossy for ultra-fine subdivisions but produces a well-formed MIDI file rather than failing).

## Repro

```python
import partitura as pt

# (1) and (2)+(3): empty <part-list/> 
pt.save_score_midi(
    pt.load_musicxml("BachJS_BWV_1006a_-_Loure.xml"),   # 0 parts
    "out.mid",
)   # before: ValueError(np.concatenate);  after: ValueError("no parts")

# (1): chord on first note
pt.load_musicxml("BachJS_Applicatio.xml")
# before: AssertionError() ;  after: loads, warns

# (4): ppq overflow
pt.save_score_midi(
    pt.load_musicxml("Ravel - Miroirs IV - Alborada del gracioso.xml"),
    "out.mid",
)   # before: struct.error;  after: warns + writes
```

## Test plan

- [x] Added `tests/test_export_robustness.py` with one regression test per fix (3 tests).
- [x] Full existing `pytest tests/` (minus the music21 test that needs an optional dep on my machine) passes — 285 passed.
- [x] Re-running our 50-piece corpus benchmark: crashes 12 → 7, all 7 remaining are the genuinely-empty-XML files now surfacing the clearer error from fix 3 rather than the numpy traceback.

Happy to split into smaller PRs if preferred.

### DIFF
--- a/partitura/io/exportmidi.py
+++ b/partitura/io/exportmidi.py
@@ -83,14 +83,9 @@ def map_to_track_channel(note_keys, mode):
 
 
 def get_ppq(parts):
-    arrays = [part.quarter_durations()[:, 1] for part in score.iter_parts(parts)]
-    if not arrays:
-        # Score has no parts (e.g., a MusicXML file with an empty <part-list/>).
-        # Caller is expected to bail out before writing notes; we just need a
-        # representable header value here so save_score_midi can produce a
-        # well-formed empty MIDI file rather than crashing in np.concatenate.
-        return 480
-    ppqs = np.concatenate(arrays)
+    ppqs = np.concatenate(
+        [part.quarter_durations()[:, 1] for part in score.iter_parts(parts)]
+    )
     ppq = int(np.lcm.reduce(ppqs))
     if ppq > SMF_MAX_PPQ:
         warnings.warn(
--- a/partitura/io/exportmidi.py
+++ b/partitura/io/exportmidi.py
@@ -4,12 +4,21 @@
 This module contains methods for exporting MIDI files
 """
 
+import warnings
+
 import numpy as np
 
 from collections import defaultdict, OrderedDict
 from typing import Optional, Iterable
 
 from mido import MidiFile, MidiTrack, Message, MetaMessage, merge_tracks
+
+# SMF type-1 / type-0 headers store ticks-per-beat in a signed 16-bit field,
+# so the largest representable ppq is 32767. Without a cap, scores with many
+# coprime divisions (e.g., 3:2 and 5:4 and 7:6 tuplets in the same piece) make
+# np.lcm.reduce explode past this limit and crash mido at write time with
+# `struct.error: 'h' format requires -32768 <= number <= 32767`.
+SMF_MAX_PPQ = 32767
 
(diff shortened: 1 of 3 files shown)


CarlosCancino-Chacon:
Hi @e7mac, thank you for the contribution! There seems to be a few merging conflicts in `partitura/io/importmusicxml.py`. Could you perhaps solve the merging conflicts? Then I will finish reviewing it.

CarlosCancino-Chacon (approved):

PR #507, opened 2026-05-20
SHUSHSHSHSHSHHHHHH

sildater:
partitura says too much, a new `quiet` flag in the import methods makes it quieter.


C-Suhit (approved):
Looks great! Just checked it and works as expected; thanks for the PR, and for the hilarious title!

Issue #506, opened 2026-05-18
Bug: Incorrect measure numbers in matchfile export for repeated segments


huispaty:
When exporting an alignment of a performance and score containing repeated segments to a matchfile, the measure numbers in the matchfile do not correctly reflect the repeat structure.

Measure numbers in the exported matchfile should match those in the score note array and measure number map.

See attached test files: Repeated notes in measures 49–56 are correctly reflected in the score note array and measure number map, but the matchfile exports incorrect measure numbers for these notes.

[match_measure_bug.zip](https://github.com/user-attachments/files/27964995/match_measure_bug.zip)

PR #508, opened 2026-05-21
Support non-unique alignments (Virtual Pointers/Sections) & fix measure numbering

darth-alexus:
This PR adds support for flexible, non-unique note alignments (as per Chiruthapudi et al., 2026) and fixes a bug with measure numbering during export.

**What’s new:**

- Virtual Pointer Notes: Added virtualSnote and virtualPnote to handle N-to-M mappings, essential for rehearsals and basso continuo realizations.

- Extended Sections: Added section and omittedSection lines for better handling of repeats and practice segments.

- Alignment Sorting: Added the _sort_alignment_ parameter to flexibly sort alignments chronologically by performance time (ptime), score time (stime), or disable sorting to preserve the original algorithm's order (False/None).

- Bug Fix (Bug: Incorrect measure numbers in matchfile export for repeated segments #506): Fixed an issue where repeated score segments resulted in incorrect measure numbers in exported match files.

**Compatibility:**
Everything should be fully backward-compatible. Version(1,0,0) match files are still supported, and new features only trigger when using the new Version(1,1,0).

_Note on create_score from match files:_
Currently, load_match(..., create_score=True) does not account for omittedSection lines. This should be done in a follow-up PR by ensuring empty bars or rests are correctly inserted to keep measure numbering consistent.

### DIFF
--- a/partitura/io/exportmatch.py
+++ b/partitura/io/exportmatch.py
@@ -473,41 +473,6 @@ def matchfile_from_alignment(
         duplicate_idx = np.hstack(list(duplicates.values()))
         voice_overlap_note_ids = list(sna[duplicate_idx]["id"])
 
-    # TODO: if version > Version(1, 0, 0): sort alignment list according to perf_time before starting the for loop
-    # Note: don't use pnote_sort_info for the sorting as time mapping probably does not make a lot of sense when sections are repeated and there are section jumps
-    # Sort alignment list according to perf_time before starting the loop for newer versions
-    
-    # should we sort the alignment at all? Shouldn't it be the alignment algorithm's job to also define where the deletions happen in the alignment?
-    if version > Version(1, 0, 0):
-        sortable_alignment = []
-        
-        for i, al in enumerate(alignment):
-            pid = al.get("performance_id")
-            
-            if pid and pid in perf_info:
-                # Matches, Insertions, Ornaments: Use exact recorded MIDI ticks
-                onset = perf_info[pid].Onset
-                sort_key = (onset, 0, i)
-            else:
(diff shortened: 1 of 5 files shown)


partitura/io/exportmatch.py
```diff
@@ -437,29 +473,84 @@ def matchfile_from_alignment(
...
+                                anchor=al_note["score_id"],
+                                attributes_list=al_note.get("score_attributes_list", []),
+                            )
+                pnote = perf_info[al_note["performance_id"]]
+                virtualSnote_note_line = MatchVirtualSnoteNote(version=version, snote=virtualSnote, note=pnote)
+                note_lines.append(virtualSnote_note_line)
+                aligned_pnotes.append(al_note["performance_id"])
+            elif al_note["score_id"] not in aligned_snotes and al_note["performance_id"] in aligned_pnotes:
+                # SNOTE - VIRTUAL_PNOTE
+                snote = score_info[al_note["score_id"]]
```
C-Suhit, line 529: I think we also need to add the score attributes list here like how it was added for the normal snote - note line? That is, I think we need to add lines 486 and 487 here as well?

`attributes_list = al_note.get("score_attributes_list", [])`
`snote.ScoreAttributesList += attributes_list`

C-Suhit: It seems to work perfectly fine otherwise, so we can merge it after resolving this!

darth-alexus: You're right, thanks! I'll updated it accordingly

C-Suhit (approved):
Works great! Thank you!

PR #513, opened 2026-08-17
Return an empty spelling for an empty note array

eeshsaxena:
`estimate_spelling` raises `IndexError` when it gets a note array with no notes, because `compute_morph_array` immediately reads `chroma_array[0]`. I ran into this loading a MIDI file that only has tempo/meta events and no notes. `estimate_key` already returns cleanly on the same empty input, so this makes spelling behave the same way and returns an empty spelling array.


CarlosCancino-Chacon:
Hi @eeshsaxena! Thanks for the PR. This is a nice and simple fix, including a meaningful test.

PR #514, opened 2026-08-17
Do not crash on a chord note that follows a rest

eeshsaxena:
When a `<note>` with a `<chord/>` follows a `<rest>`, `_handle_note` reads `prev_note.is_grace_chord`, but `prev_note` is a `Rest` and only `Note` carries that attribute, so the whole load dies with `AttributeError`. The existing code already treats a `<chord/>` with no preceding note (prev_note is None) as malformed and continues with a warning, so I widened that same guard to cover a rest as well. Added a robustness test next to the first-note one.

### DIFF
--- a/partitura/io/importmusicxml.py
+++ b/partitura/io/importmusicxml.py
@@ -1371,14 +1371,14 @@ def _handle_note(e, position, part, ongoing, prev_note, doc_order, prev_beam=Non
     if chord is not None:
         # this note starts at the same position as the previous note, and has
         # same duration
-        if not isinstance(prev_note, score.Note):
-            # Malformed MusicXML: a <chord> child with no note to anchor to,
-            # either the first note of a part (or first after a backup/forward),
-            # where prev_note is None, or one that follows a <rest>, where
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Do not crash on a chord note that follows a rest`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining partitura!

sildater:
Hi @eeshsaxena thank you very much for the contribution! It looks good to me, the only edge case I see is that unpitched notes (for percussion etc) may also be wrapped by chord tags. In partitura they are a separate class `UnpitchedNote` , could you modify the instance check to capture these as well? Thank you!

eeshsaxena:
Good catch, thanks. Unpitched percussion notes should definitely be covered too. I'll widen the isinstance check to include UnpitchedNote and push an update shortly.

sildater (approved):

sildater:
thanks for the PR, the changes look good to me!

